PR #4, opened 2016-09-05
Fixed .gitattributes for linguist language detection

approxit:
Obj-c is evil, c++ is overhead.
Stats are crusial, let's make them accurate!


tehKaiN:
crucial*

PR #21, opened 2017-11-04
Simplified casting in blitter pointers

tehKaiN:
Related to our discussion about C casting syntax. Some parentheses weren't necessary.


tehKaiN:
This gets quite old, so I guess I'll merge it.

Issue #24, opened 2017-11-18
Improve mouse manager


tehKaiN:
Currently, mouse actions are collected using Intuition events. Also, current cursor codebase relies on OS stuff, which is conflicting with current custom copperlist implementation. There is a need to:

- [ ] ~~Modernize API by integrating cursor into copperlist,~~
- [x] Add ability to have 2 cursors on screen (vide Settlers split-screen mode),
- [ ] ~~Display as 3bpp or 4bpp sprite with configurable sprite channel index (indices),~~
- [x] Display as the only thing on sprite channel
- [x] Display as part of sprite list for channel
- [x] Grab cursor events directly for hardware - needed in #13

tehKaiN:
Dual mice position and 3 button states support through hardware has been implemented on branch `hardware-input`. Display is still not done.

deuteros76:
Is there any example of how to use a mouse? 

I'm looking at Open Fire's code. I've copied cursor.c, cursor.h and cursor.bm to my project and then I included the cursorCreate, cursorUpdate and cursorDestroy calls but nothing happens. 

Maybe I missed something related to the Copper. Anyway it would be good to have an example of a mouse application in the showcase.

tehKaiN:
Using mouse can be done with mouse manager. Displaying cursor can be done by sprite manager.

Until documentation is produced in #120, example code can be extracted from https://github.com/Last-Minute-Creations/slipgates and https://github.com/Last-Minute-Creations/openFire/ . 

Displaying 4bpp cursor will be implemented in #212.

Issue #13, opened 2017-10-26
Separate keyboard & mouse processing


tehKaiN:
Currently processing mouse & keyboard are bound together by Intuition event loop. This prevents isolation of these two managers.

I guess the best way to do this would be getting mouse & keyboard events directly from HW, omitting Intuition code.

tehKaiN:
Done in `hardware-input` branch.

PR #29, opened 2017-12-18
Hardware keyboard & mice input

tehKaiN:
This pull requests closes pull request #13 and does some work for #24 . Also, it brings as closer to abandoning intuition.

### DIFF
--- a/include/ace/managers/mouse.h
+++ b/include/ace/managers/mouse.h
@@ -126,7 +126,7 @@ static inline void mouseSetButton(
  * Returns given mouse's button state.
  * @param ubMousePort: Mouse to be polled.
  * @param ubMouseCode: Button to be polled (MOUSE_LMB, MOUSE_RMB or MOUSE_MMB).
- * @return Current mouse sate: (MOUSE_NACTIVE, MOUSE_USED, MOUSE_ACTIVE).
+ * @return 1 if button is pressed, otherwise 0.
  */
 static inline UBYTE mouseCheck(
 	IN UBYTE ubMousePort,
--- a/include/ace/managers/mouse.h
+++ b/include/ace/managers/mouse.h
@@ -2,17 +2,12 @@
 #define GUARD_ACE_MANAGER_MOUSE_H
 
 #include <ace/types.h>
+#include <ace/macros.h>
 
 /* Types */
-#ifdef AMIGA
-#define MOUSE_LMB IECODE_LBUTTON
-#define MOUSE_RMB IECODE_RBUTTON
-#define MOUSE_MMB IECODE_MBUTTON
-#else
-#define MOUSE_LMB 1
-#define MOUSE_RMB 2
-#define MOUSE_MMB 4
-#endif // AMIGA
+#define MOUSE_LMB 0
+#define MOUSE_RMB 1
+#define MOUSE_MMB 2
 
 #define MOUSE_NACTIVE 0
 #define MOUSE_USED 1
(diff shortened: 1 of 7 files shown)


include/ace/managers/mouse.h
```diff
@@ -1,101 +1,222 @@
...
-	IN UWORD uwWidth,
-	IN UWORD uwHeight
+/**
+ * Set on-screen constraints for cursor.
+ * @param uwLoX Minimum cursor X position.
+ * @param uwLoX Minimum cursor Y position.
+ * @param uwHiY Maximum cursor X position.
+ * @param uwHiY Maximum cursor Y position.
+ */
+void mouseSetBounds(
```
approxit: When we have support for "multiple" simultaneous mouse handling... We should be able to set bounds for specific mouse.

Think about possibility to make split screen games with 2 connected mouses.

include/ace/utils/custom.h
```diff
@@ -13,6 +13,11 @@
 // Here was __far attrib from DICE/GCC times. Not needed for VBCC, so it was removed.
 extern struct Custom custom;
 
+/**
+ * Bit value macro - useful for setting & testing bits.
+ */
+#define BV(x) (1 << (x))
```
approxit: Should this be moved to more generic file?

src/ace/managers/mouse.c
```diff
@@ -1,126 +1,128 @@
...
-#endif // AMIGA
-}
+	volatile UWORD * const pCiaAPra = (UWORD*)((UBYTE*)0xBFE001);
+
+	// Deltas are signed bytes even though underflows and overflows may occur.
+	// It is expected behavior since it is encouraged in Amiga HRM as means to
+	// determine mouse movement direction which takes into account joyxdat
+	// underflows and overflows.
+
+	if(g_sMouseManager.ubPortFlags & MOUSE_PORT_1) {
```
approxit: This IFs should be totally a function with few variables. Only 4 hardcoded elements varies.

src/ace/managers/mouse.c
```diff
@@ -1,126 +1,128 @@
...
-		g_sMouseManager.pStates[ubMouseCode] = MOUSE_USED;
-		return 1;
+	mouseSetBounds(0, 0, SCREEN_PAL_WIDTH, SCREEN_PAL_HEIGHT);
+
+	g_sMouseManager.ubPortFlags = ubPortFlags;
+	g_sMouseManager.uwPrevPotGo = custom.potinp;
+	UWORD uwPotMask = 0;
+
+	// Enable RMB & MMB
+	if(ubPortFlags & MOUSE_PORT_1) {
```
approxit: This IFs should be totally a function with few variables. Only 1 hardcoded element varies.

tehKaiN: Not really, since there is no need for that. Also, `mouseCreate()` is short enough to have this stuff written as it is right now. No need for moving every repetitive code to separate fn.

include/ace/utils/custom.h
tehKaiN: Done in 7ba50f5

src/ace/managers/mouse.c
tehKaiN: Done in dab7e1d

include/ace/managers/mouse.h
tehKaiN: Done in 776fae8

Issue #41, opened 2018-01-12
Add optimizable rotate fns


tehKaiN:
Since m68k has barrel-shift functions, it is possible to have rotate fns.

https://stackoverflow.com/questions/776508/best-practices-for-circular-shift-rotate-operations-in-c these are supposed to be optimizable to single instruction on decent compilers, however this should be checked on VBCC and Bebbo's GCC.

tehKaiN:
Functions added in #42. Not sure if they will optimize away nicely, so this should be checked and reported to Bebbo and/or VBCC core & amiga target authors.

Issue #40, opened 2018-01-12
Add line drawing functions


tehKaiN:
Currently, there are no means to draw lines using blitter. Drawing vertical/horizontal lines using `blitRect()` is probably not fastest possible solution, and most certailny such usage is not intuitive.

Implementation could be based on [Cahir's routine](https://github.com/cahirwpz/demoscene/blob/master/a500/base/libsys/blt-line.c).

- [x] Add blitLine() or something
- [ ] fix bugs

tehKaiN:
Added blitLine() based on Cahir's function in #42. There is still one bug: on some angles, lines drawn on different bitplanes don't overlap perfectly, resulting in color inconsistency.

PR #42, opened 2018-01-20
Blit lines and rotate fns

tehKaiN:
Related to #41 and #40. Not sure if those rotate fns will get optimized away in Bebbo's GCC but if not, I'll start an issue there. ;)


approxit:
LGTM

PR #49, opened 2018-03-16
Compile gcc

tehKaiN:
Squared and OpenFire are building fine with GCC.
vasm include path is temporary measure until pario gets rewritten to C and there will be no asm dependency.


showcase/src/menu/menu.c
```diff
@@ -108,11 +108,11 @@ void gsMenuDestroy(void) {
 
 void menuDrawBG() {
 	UWORD uwX, uwY;
-	UBYTE ubOdd, ubColor;
+	UBYTE ubOdd = 0, ubColor;
```
approxit: This line could be actually tricky. I'm not a big fan of declarations multiple variables in one line. Highlighted line can be error prone when assignment would be after ubColor. Someone can think of assigning same value to all declarations, which is not true.

Would totally go for single declaration per line.

showcase/src/test/font.c
```diff
@@ -84,7 +84,8 @@ void gsTestFontTableLoop(void) {
 
 void gsTestFontSentenceLoop(void) {
 	UBYTE i, d, ubRedraw;
-	static const UBYTE szAllowedChars[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+	static const char szAllowedChars[] =
```
approxit: 👍

showcase/src/menu/menu.c
tehKaiN: Fixed in 52602be

src/ace/managers/key.c
```diff
@@ -52,8 +54,8 @@ void keyCreate(void) {
 	g_sKeyManager.pInt->is_Node.ln_Type = NT_INTERRUPT;
 	g_sKeyManager.pInt->is_Node.ln_Pri = -60;
 	g_sKeyManager.pInt->is_Node.ln_Name = "ACE_Keyboard_CIA";
-	g_sKeyManager.pInt->is_Data = (APTR)&g_sKeyManager;
-	g_sKeyManager.pInt->is_Code = keyIntServer;
+	g_sKeyManager.pInt->is_Data = 0;
+	g_sKeyManager.pInt->is_Code = (void(*)(void))keyIntServer;
```
approxit: What's the background of this change?

src/ace/managers/log.c
```diff
@@ -165,17 +165,17 @@ void _logAvgEnd(tAvg *pAvg) {
 		pAvg->ulMin = pAvg->pDeltas[pAvg->uwCurrDelta];
 	++pAvg->uwCurrDelta;
 	// Roll
-	if(pAvg->uwCurrDelta == pAvg->uwAllocCount)
+	if(pAvg->uwCurrDelta >= pAvg->uwAllocCount) {
```
approxit: In every project I'm involved I'm always prefer to use `<` and `<=` operators. Making this consistent enables faster reading as increasing X value on typical chart from math classes.

src/ace/managers/key.c
tehKaiN: Bebbo GCC's OS headers had at some point wrong interrupt struct decls, which were causing warnings for interrupt servers with custom argument list. Cast is not needed now. Fixed in 4296300.

Regarding data, this field contains value which is passed as interrupt handler's arg. Not needed here since manager is global and inside handler there is a function call which internally uses global manager.

src/ace/managers/log.c
tehKaiN: I prefer typing it in a way that's naturally coming to your mind. You'd rather say that "var has reached or passed its limit", than a "limit is equal or smaller than var".

src/fixmath/fix16_trig.c
```diff
@@ -16,13 +16,13 @@ static fix16_t _fix16_atan_cache_value[4096] = { 0 };
 
 fix16_t fix16_sin_parabola(fix16_t inAngle)
 {
-	fix16_t abs_inAngle, abs_retval, retval;
+	fix16_t abs_inAngle, retval;
```
approxit: `abs_` variable name prefix?

tehKaiN: That's not mine, but fixmath's. I've moved abs_retval past ifdef so that gcc would not yell at me that I have unused variable.

PR #51, opened 2018-04-28
Os disabling

### LINKED ISSUE #31: Add ability to take over whole HW
To achieve best performance, it is crucial to be able to disable OS stuff completely, including multitasking. After removing Intuition-related code, there should be no blockers to achieve this.

- [x] Implement OS disabling and re-enabling
- [x] Force OS enabling for FS functions - needs #25 
- [ ] Make interrupt servers also register in OS so that keyboard is properly updated when OS is active - needed for https://github.com/tehKaiN/ld40-squared/issues/1

tehKaiN:
Fixes #31 
Last point would be dealt with later.


tehKaiN:
@approxit has reviewed this but was too lazy to click merge btn. Let's do this right now.

PR #53, opened 2018-04-28
"Meta" milestone

### LINKED ISSUE #14: GAME_DEBUG define as compile flag
Right now we have GAME_DEBUG flag hardcoded inside source files.
It's changes results in git. To make a prod you need to switch it's value on and off.

Better idea is to move this flag to compile flags.

### LINKED ISSUE #43: Determine license
There are some options. MIT, BSD, Unlicense or GPL. Also, this decision may affect our prods' licenses.

## Viral vs Non-viral (GPL)

Pros:

- It would be cool to have more and more open source stuff on Amigas. GPL would enforce it.

Cons:

- People may prefer using other dev tools which don't enforce opening their sources.
- There are some ppl in retrodev who'd like to keep their sources closed
- KaiN hates those mile-long comments on top of every source file
- If we get other contributors that would mean tracing fn history in their respective comment blocks, so that each contributor gets credit for their work

## Attribution vs non-attribution (BSD)

Pros:

- Enforcing some kind of attribution in prod could bring more popularity to project

Cons:

- Again, not everyone may be happy with such enforcement and such enforcement could play against us

## MIT vs Unlicense

All cons mentioned above are irrelevant if we just use MIT/Unlicense.

MIT is just an info about author without any restrictions, so It's mostly cool. What isn't cool is that most licenses require list of authors who modified source files. I don't really like keeping up-to-date list of contributors in copyright notice. That's getting messy very quickly.

So I propose using unlicense - at first line text below becomes copyrighted, then all rights are passed to everyone who's reading. Since there are no rights to be cleaimed, it's irrelevant who's listed as author. And there are really more important things to do than looking after all that license bs.

### LINKED ISSUE #48: Move util/manager-specific logging functions from log.c
- [x] logBitMap
- [x] logUCopList

tehKaiN:
fixes #43 
fixes #14 
fixes #48

### DIFF
--- a/showcase/src/test/copper.c
+++ b/showcase/src/test/copper.c
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 #include "test/blit.h"
 
 #include <ace/utils/extview.h>
--- a/showcase/src/test/copper.c
+++ b/showcase/src/test/copper.c
@@ -109,16 +109,16 @@ void gsTestCopperCreate(void) {
 		TEST_COPPER_COLOR_BORDER, 0xFFFF, 0
 	);
 
-	for(i = 0; i != 32; ++i) {
+	for(i = 0; i < 32; ++i) {
 		pBar[i] = copBlockCreate(s_pTestCopperView->pCopList, 1, 0, 50+i);
 	}
-	for(i = 0; i != 16; ++i) {
+	for(i = 0; i < 16; ++i) {
 		copMove(
 			s_pTestCopperView->pCopList, pBar[i],
 			&g_pCustom->color[1], colorHSV(0,255,i << 3)
 		);
 	}
-	for(i = 16; i != 32; ++i) {
+	for(i = 16; i < 32; ++i) {
 		copMove(
 			s_pTestCopperView->pCopList, pBar[i],
 			&g_pCustom->color[1], colorHSV(0,255,(31-i) << 3)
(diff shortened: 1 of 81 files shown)


showcase/src/test/copper.c
```diff
@@ -136,19 +141,24 @@ void gsTestCopperLoop(void) {
...
-	if(uwY <= 30)
+	}
+	if(uwY <= 30) {
 		bDir = 1;
+	}
 
 	uwY += 2*bDir;
 
-	for(i = 0; i != 32; ++i)
+	for(i = 0; i != 32; ++i) {
```
approxit: `!=` instead of `<=`?

src/ace/managers/timer.c
```diff
@@ -162,5 +171,5 @@ void timerWaitUs(UWORD uwUsCnt) {
 	// timerGetPrec(): One tick equals: PAL - 0.40us, NTSC - 0.45us
 	ULONG ulStart = timerGetPrec();
 	UWORD uwTickCnt = uwUsCnt*2/5;
-	while(timerGetPrec() - ulStart < uwTickCnt);
+	while(timerGetPrec() - ulStart < uwTickCnt) {}
```
approxit: Good example for clangformat. Guess adding something to empty code blocks will show in diff, which we agreed to mitigate. Should be in new lines?

src/ace/managers/viewport/scrollbuffer.c
```diff
@@ -31,10 +35,15 @@ tScrollBufferManager *scrollBufferCreate(tVPort *pVPort, UBYTE ubMarginWidth, UW
 	vPortAddManager(pVPort, (tVpManager*)pManager);
 
 	// Find camera manager, create if not exists
-	if(!(pManager->pCameraManager = (tCameraManager*)vPortGetManager(pVPort, VPM_CAMERA)))
-		pManager->pCameraManager = cameraCreate(pVPort, 0, 0, uwBoundWidth, uwBoundHeight);
-	else
+	pManager->pCameraManager = (tCameraManager*)vPortGetManager(pVPort, VPM_CAMERA);
+	if(!pManager) {
```
approxit: Should it be `pManager->pCameraManager`?

src/ace/managers/timer.c
tehKaiN: I think that such change is significant enough to have it fully shown in diff. Also, the following code looks awful and doesn't look as done deliberately.

``` c
while(timerGetPrec() - ulStart < uwTickCnt) {
}
```

src/ace/managers/viewport/scrollbuffer.c
tehKaiN: Thanks!

showcase/src/test/copper.c
tehKaiN: Thanks!

PR #58, opened 2018-05-06
Dbl bfr

### LINKED ISSUE #34: Dust off scroll buffer & tile buffer managers
I guess this part of code is bugged and rotten since it wasn't touched for a long time.

needed for https://github.com/tehKaiN/openFire/issues/5

Scroll buffer:

- [x] Migrate to taglist syntax
- [x] On scroll buffer, check if bobs are drawing correctly where bitmap folds (interleaved)

Tile buffer:

- [x] Migrate to taglist syntax
- [x] Fix initial draw
- [x] check if tiles are redrawing properly on draw margins (checked Y for AMiner)
- [x] fix garbage which appears during x-movement on Y wrap line - wasn't ACE's fault

### LINKED ISSUE #57: Implement double buffering in simple buffer manager
Double buffering is needed for drawing if you don't want to race the display beam.

Many years ago I had some clever idea of implementing double buffering as vport manager which could work regardless if one uses simple or scroll buffer, but right now I have no idea how that could work. The only remnant of this idea is in viewport manager priority list:

``` c
/**
 *  Viewport manager IDs.
 *  Number determines processing order. Camera is last, so rest may see
 *  a difference between its current and previous position
 */
#define VPM_SCROLL       0
#define VPM_TILEBUFFER   1
#define VPM_DOUBLEBUFFER 2
#define VPM_CAMERA       128
```

### LINKED ISSUE #59: Implement double buffering in tile & scroll buffer managers
Could be done in similar way as in #57

tehKaiN:
fixes #57 
fixes #34 
fixes #59 

Imposes a breaking change: pointer to simplebuffer's bitmap is changed from `pManager->pBuffer` to `pManager->pFront` and `pManager->pBack`.

### DIFF
--- a/src/ace/managers/viewport/scrollbuffer.c
+++ b/src/ace/managers/viewport/scrollbuffer.c
@@ -3,33 +3,66 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include <ace/managers/viewport/scrollbuffer.h>
+#include <ace/utils/tag.h>
+#include <limits.h>
+
 
 #ifdef AMIGA
 
-/**
- * Creates scroll manager
- */
-tScrollBufferManager *scrollBufferCreate(tVPort *pVPort, UBYTE ubMarginWidth, UWORD uwBoundWidth, UWORD uwBoundHeight) {
-	logBlockBegin("scrollBufferCreate(pVPort: %p, ubMarginWidth: %hu, uwBoundWidth: %u, uwBoundHeight: %u)", pVPort, ubMarginWidth, uwBoundWidth, uwBoundHeight);
+tScrollBufferManager *scrollBufferCreate(void *pTags, ...) {
+	va_list vaTags;
+	tCopList *pCopList;
 	tScrollBufferManager *pManager;
+	UBYTE ubMarginWidth;
+	UWORD uwBoundWidth, uwBoundHeight;
+	UBYTE ubBitmapFlags;
+	UBYTE isCameraCreated = 0;
+	UBYTE isDblBfr;
 
-	// Wype�nienie struktury mened�era
-	pManager = memAllocFast(sizeof(tScrollBufferManager));
-	logWrite("Addr: %p\n", pManager);
-	pManager->sCommon.pNext = 0;
+	logBlockBegin("scrollBufferCreate(pTags: %p, ...)", pTags);
+	va_start(vaTags, pTags);
+
+	// Init manager
+	pManager = memAllocFastClear(
+		sizeof(tScrollBufferManager)
+	);
 	pManager->sCommon.process = (tVpManagerFn)scrollBufferProcess;
 	pManager->sCommon.destroy = (tVpManagerFn)scrollBufferDestroy;
 	pManager->sCommon.ubId = VPM_SCROLL;
+	logWrite("Addr: %p\n", pManager);
+
+	tVPort *pVPort = (tVPort*)tagGet(pTags, vaTags, TAG_SCROLLBUFFER_VPORT, 0);
+	if(!pVPort) {
+		logWrite("ERR: No parent viewport (TAG_SCROLLBUFFER_VPORT) specified!\n");
+		goto fail;
+	}
 	pManager->sCommon.pVPort = pVPort;
-	pManager->pBuffer = 0;
-	pManager->uwModulo = 0;
-	pManager->pAvg = logAvgCreate("scrollBufferProcess()", 100);
+	logWrite("Parent VPort: %p\n", pVPort);
 
-	scrollBufferReset(pManager, ubMarginWidth, uwBoundWidth, uwBoundHeight);
+	ubMarginWidth = tagGet(
+		pTags, vaTags, TAG_SCROLLBUFFER_MARGIN_WIDTH, UCHAR_MAX
+	);
+	if(ubMarginWidth == UCHAR_MAX) {
+		logWrite(
+			"ERR: No margin width (TAG_SCROLLBUFFER_MARGIN_WIDTH) specified!\n"
+		);
+		goto fail;
+	}
 
-	// Create copperlist
-	pManager->pStartBlock = copBlockCreate(pVPort->pView->pCopList, 2*pVPort->ubBPP + 8, 0, 0);
-	pManager->pBreakBlock = copBlockCreate(pVPort->pView->pCopList, 2*pVPort->ubBPP + 2, 0x7F, 0xFF);
+	// Buffer bitmap
+	uwBoundWidth = tagGet(
+		pTags, vaTags, TAG_SCROLLBUFFER_BOUND_WIDTH, pVPort->uwWidth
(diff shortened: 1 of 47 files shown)


src/ace/managers/viewport/simplebuffer.c
```diff
@@ -39,18 +131,31 @@ tSimpleBufferManager *simpleBufferCreate(
...
 		goto fail;
 	}
 
+	UBYTE isDblBfr = tagGet(pTags, vaTags, TAG_SIMPLEBUFFER_IS_DBLBUF, 0);
+	if(isDblBfr) {
+		pBack = bitmapCreate(
+			uwBoundWidth, uwBoundHeight,
+			pVPort->ubBPP, ubBitmapFlags
+		);
+		if(!pFront) {
```
approxit: `pFront`?

src/ace/managers/viewport/simplebuffer.c
```diff
@@ -81,7 +186,8 @@ tSimpleBufferManager *simpleBufferCreate(
 		logWrite("Copperlist offset: %u\n", pManager->uwCopperOffset);
 	}
 
-	simpleBufferSetBitmap(pManager, pBuffer);
+	simpleBufferSetFront(pManager, pFront);
+	simpleBufferSetBack(pManager, pBack == 0 ? pFront : pBack);
```
approxit: Explicit `== 0` is not needed.
`pBack ? pBack : pFront`

src/ace/managers/viewport/simplebuffer.c
```diff
@@ -90,89 +196,30 @@ tSimpleBufferManager *simpleBufferCreate(
...
-			copSetMove(&pCmdList[6 + i*2 + 1].sMove, &g_pBplFetch[i].uwLo, ulPlaneAddr & 0xFFFF);
-		}
-		// Copy to front buffer since it needs initialization there too
-		CopyMem(
-			&pCopList->pBackBfr->pList[pManager->uwCopperOffset],
-			&pCopList->pFrontBfr->pList[pManager->uwCopperOffset],
-			(6+2*pManager->sCommon.pVPort->ubBPP)*sizeof(tCopCmd)
-		);
+	if(pManager->pCameraManager && isCameraCreated) {
+		cameraDestroy(pManager->pCameraManager);
```
approxit: Accessing `pManager` that was freed in line 206?

approxit:
I need to say **NO** for breaking change with removing `pBuffer` pointer.

Actually games explicitly use `pBuffer` everywhere. `pBack/pFront` is a hard coded pattern bound only to simplebuffer manager. Other buffer managers use `pBuffer`. You don't want to refactor whole game code while switching on/from simplebuffer manager.

Let's keep the old pointer to keep sanity, and `pFront/pBack` as additional pointers.

tehKaiN:
Okay, hold on. This braking change is minor and I've mentioned it only so that you can effortlessly update code in your prods. Keep in mind that we're not versioning ACE yet, so unless we're in semver 1.0 expect stuff to change a lot.

I expect that ACE will have little to no backwards compatibility between versions for sake of speed, because there are lots of stuff in code which were written in a very poor way. Best approaches are not yet known, so I'm not going to support wrong ones.

This change is for the better, since for single-buffered screens `pFront` and `pBack` will be the same, and if you work on `pBack` then you can switch to double buffered stuff without changing anything else except for one creation tag.

Also, recent OS disabling and GCC pull requests probably broke your code anyway, so idk why you're getting mad at such trivial change.

If you really don't want to change code in your games, just include ACE as submodule with HEAD set to last known commit which worked.

src/ace/managers/viewport/simplebuffer.c
```diff
@@ -187,7 +187,7 @@ tSimpleBufferManager *simpleBufferCreate(
 	}
 
 	simpleBufferSetFront(pManager, pFront);
-	simpleBufferSetBack(pManager, pBack == 0 ? pFront : pBack);
+	simpleBufferSetBack(pManager, pBack ? pFront : pBack);
```
approxit: Ommiting `== 0` actually flips the logic.

tehKaiN: fixed in 6301206

tehKaiN:
We may wait with merge until rest of vport managers get double buffering support then - it will be consistent across whole engine.

From my current point of view, writing games using single buffering shouldn't be a default approach since it works in very few cases, where you're sure you will always be with your drawing before display beam. Double buffering allows you to not waste computing time on bob sorting which gets costly if you also consider z-sort. Without sorting in mind, your only concern is if you finish all blits in time of single frame, or in two frames if you target 25fps instead of 50.

src/ace/managers/viewport/scrollbuffer.c
```diff
@@ -148,40 +226,66 @@ void scrollBufferReset(tScrollBufferManager *pManager, UBYTE ubMarginWidth, UWOR
...
 	pManager->uBfrBounds.sUwCoord.uwY = uwBoundHeight;
-	pManager->uwBmAvailHeight = ubMarginWidth * blockCountCeil(uwVpHeight, ubMarginWidth) + ubMarginWidth*4;
+	// TODO optimize avail height to power of two so that modulo can be an AND
+	pManager->uwBmAvailHeight = ubMarginWidth * (blockCountCeil(uwVpHeight, ubMarginWidth) + 4);
 
 	// Destroy old buffer bitmap
-	if(pManager->pBuffer) {
-		bitmapDestroy(pManager->pBuffer);
+	if(pManager->pFront != pManager->pBack) {
+		bitmapDestroy(pManager->pFront);
```
approxit: Can `pFront` be zero?

tehKaiN: shouldn't be, but I'll change it just to be sure

src/ace/managers/viewport/simplebuffer.c
```diff
@@ -90,151 +200,98 @@ tSimpleBufferManager *simpleBufferCreate(
...
+	logBlockEnd("simpleBufferCreate()");
+	va_end(vaTags);
+	return 0;
 }
 
 void simpleBufferDestroy(tSimpleBufferManager *pManager) {
 	logBlockBegin("simpleBufferDestroy()");
 	logWrite("Destroying bitmap...\n");
-	bitmapDestroy(pManager->pBuffer);
+	if(pManager->pBack != pManager->pFront) {
```
approxit: Can `pBack` be zero too?

tehKaiN: shouldn't be - in `simpleBufferCreate()` `pFront` gets created and if dbl bfr then `pBack` gets allocated, otherwise it's set to `pFront`. So freeing `pBack` should only take place when there's dlb buffering. Unless users is an idiot which pokes with code too much.

On the second thought I wonder if I should've added safety in scroll buffer.

tehKaiN:
i don't want to wait for it any longer, seems stable enough

PR #65, opened 2018-06-24
Added memory trash detection

tehKaiN:
Eases out-of-bounds debugging in heap-allocated regions.

Right trashing occurs mostly when array indices become bigger than array size. Left trashes are more tricky to interpret, but are indicating that something's wrong is going on in memory.

### DIFF
--- a/src/ace/managers/memory.c
+++ b/src/ace/managers/memory.c
@@ -144,7 +144,7 @@ void _memDestroy(void) {
 
 void *_memAllocDbg(ULONG ulSize, ULONG ulFlags, UWORD uwLine, char *szFile) {
 	void *pAddr;
-	pAddr = _memAllocRls(ulSize, ulFlags);
+	pAddr = _memAllocRls(ulSize + 2*sizeof(ULONG), ulFlags);
 	if(!pAddr) {
 		filePrintf(
 			s_pMemLog, "ERR: couldn't allocate %lu bytes! (%s:%u)\n",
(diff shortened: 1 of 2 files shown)


src/ace/managers/memory.c
```diff
@@ -164,13 +164,19 @@ void *_memAllocDbg(ULONG ulSize, ULONG ulFlags, UWORD uwLine, char *szFile) {
 #endif // AMIGA
 		return 0;
 	}
+	pAddr += sizeof(ULONG);
+	ULONG *pCafe = (ULONG*)(pAddr - sizeof(ULONG));
+	ULONG *pDead = (ULONG*)(pAddr + ulSize);
+	*pCafe = 0xCAFEBABE;
+	*pDead = 0xDEADBEEF;
```
approxit: 168-171 are useless?

tehKaiN: wtf mate, that's the whole point of this pull request - you set up canaries on start/end of memory alloc and during freeing it you check if it's intact

src/ace/managers/memory.c
```diff
@@ -144,7 +144,7 @@ void _memDestroy(void) {
 
 void *_memAllocDbg(ULONG ulSize, ULONG ulFlags, UWORD uwLine, char *szFile) {
 	void *pAddr;
-	pAddr = _memAllocRls(ulSize, ulFlags);
+	pAddr = _memAllocRls(ulSize + 2*sizeof(ULONG), ulFlags);
```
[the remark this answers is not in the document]
tehKaiN: that's not something we agreed upon in our last codestyle issue - we'll have this discussion somewhere else at the later time. I'll add missing spaces instead since IMO brackets would decrease code readability

PR #66, opened 2018-07-09
Directory creation fns

tehKaiN:


include/ace/utils/dir.h
```diff
@@ -48,3 +48,42 @@ UBYTE dirRead(tDir *pDir, char *szFileName, UWORD uwFileNameMax);
...
+ * This function accepts paths, but will expect that only top-most directory
+ * needs to be created.
+ *
+ * @param szName Path to directory to be created.
+ * @return Success: 1, otherwise 0.
+ *
+ * @see dirExists()
+ * @see dirCreatePath()
+ */
+UBYTE dirCreate(const char *szName);
```
approxit: I don't get one thing. If `dirCreate` can make a path, why where is a `dirCreatePath` as a separate function? The top-most-dir-must-exist limitation is too little for me to have function that create `n-1` levels, and `n` levels of given path.

tehKaiN: Try reading doxycomment again ;)

PR #79, opened 2019-01-11
Tools/font conv

### LINKED ISSUE #44: Add font generator utility
Currently, I have php script for generating ACE fonts, but it's inconvenient and not portable way to do this. It needs C implementation without too many dependencies - should be working directly on ttf/otf and some popular bitmap font formats.

Perhaps it would be wise to use freetype amalgamated lib?

---

## TODO

- [x] Generate 1bpp bitmap glyphs, each in separate PNG file, based on TTF
- [x] Generate ACE packed font in PNG file
- [x] Organize app so that it can use either TTF or PNGs as source
- [x] implement `glyphsFromDir()`
- [x] Generate ACE packed font
- [x] Add parameters for CLI & help screen

tehKaiN:
closes #44 

@approxit you have 24hrs for review, then it's going to master since it's new code anyway


tehKaiN:
this had too big scope when branched out of dbl-bfr, but it's fixed now.

PR #87, opened 2019-02-22
Tools/tileset conv

### LINKED ISSUE #83: Add tool to convert tilesets
ACE uses tilesets organized in single column - this way only tile's Y offset needs to be calculated, X is always 0. This makes Y calc is easy as in left shifting by tile size. Unfortunately composing tileset image in such manner is cumbersome. A tool is needed to automate such task.

When dealing with some irregular tile source bitmaps (e.g. tiles drawn on big PNG on arbitrary coords) one could use `bitmap_extract` tool to generate lots of separate images of single tiles and then merge them into ACE's tileset format.

- [x] Implement slicing single-column PNG tileset into directory with each tile held separately and named using tile number
- [x] Implement merging directory of index-named tile graphics into single-column PNG / ACE BM

tehKaiN:
fixes #83


tehKaiN:
@approxit ?

PR #89, opened 2019-03-01
Fix/cop block disable

### LINKED ISSUE #88: Problems on AGA machines
When following current tutorial up to view/viewport portion, palette setting may not work on AGA machines. This is probably caused by different bplcon3 states.

A fix may be as simple as doing `g_pCustom->bplcon3 = 0;` along with fmode in `viewLoad()` fn.

I need some help here since I don't own AGA machine.

tehKaiN:
Minor fixes for scrollBuffer.

Also fixes #88


tehKaiN:
Ozzyboshi has tested AGA fix, so I'm merging it without review.

PR #93, opened 2019-03-25
Bitmap fastmem

tehKaiN:
This allows having some gfx data loaded to FAST and then memcpy'd into CHIP when needed.


tehKaiN:
@approxit I'll merge this 'cause I need this

PR #99, opened 2019-05-17
Fix comparison issue con tPlanarBitmap constructor

Ozzyboshi:
If palette size is 256 the uint_t counter on tPlanarBitmap constructor is not enough and we end up on a endless loop because the comparison will always be performed with something less of 256 (unsigned int g
    I changed added 8 more bits to the i counter and casted Palette.m_vColors.size for clarity, in this way 16 bits will be involved on the comparison and they are enough for 256 colors.


tehKaiN:
nice, thanks!

PR #101, opened 2019-05-27
fix palette conv endless loop if source file does not end with a newline

Ozzyboshi:
Today I had an issue with palette_conv,

if the input file does not end with an empty line, palette_conv is stuck in an endless loop.
My take on this is if the end of file is reached, the szLine is not cleared by std:getline and the exit condition is never met ( szLine == "")

My input file was not generated with gimp, it was a simple copy and paste from another text file, but It had the same structure of yours except for the last line.

Please review my additions, tell me what you think about this issue and eventually merge it into master.

### DIFF
--- a/tools/src/common/palette.cpp
+++ b/tools/src/common/palette.cpp
@@ -36,6 +36,7 @@ tPalette tPalette::fromGpl(const std::string &szPath)
 
 	// Read colors
 	bool isEnd = false;
+	bool isNextEnd = false;
 	do {
 		std::stringstream ss(szLine);
 		int r, g, b;
@@ -46,9 +47,11 @@ tPalette tPalette::fromGpl(const std::string &szPath)
 		Palette.m_vColors.push_back(Color);
 
 		std::getline(Source, szLine);
-		if(szLine == "") {
+		if(isNextEnd || szLine == "") {
 			isEnd = true;
 		}
+		if(Source.eof())
+			isNextEnd=true;
 	} while(!isEnd);
 
 	fmt::print("Palette color count: {}\n", Palette.m_vColors.size());


tehKaiN:
Thanks! Of course this has to be fixed and it would be good enough, but I wonder if it's possible to not introduce another bool. I wonder if following code would work.

``` cpp
 	do {
		if(szLine != "") {
			std::stringstream ss(szLine);
			int r, g, b;
			ss >> r;
			ss >> g;
			ss >> b;
			Color = tRgb(r, g, b);
			Palette.m_vColors.push_back(Color);
		}
	} while(!std::getline(Source, szLine).eof()); 
```

If not, everything in your code is ok except for `nextEnd` name. I haven't documented it yet in [codestyle](https://github.com/AmigaPorts/ACE/blob/master/docs/contributing/codestyle.md), but I try to prefix boolean vars with one of following: `is`, `has`, `was`. So in our case it should be refactored to `wasEof` or something.

Since you are the author of the fix, I'll leave a choice to you - do you want to merge your fix or are we sticking to my (yet untested) code? ;)

Ozzyboshi:
I gave a try with your code, it works, in this way palette_conv accepts input file without empty line at the end, however, in this case, the output file is missing 2 bytes, probably your "do while" loop does not perform the last iteration.
I would stick with my own fix, right now I can't come up with another solution without a new boolean variable.
For the naming convention issue you are right, i will call the variable isNextEnd if you are good with the extra variable.
Please let me know what you can find another solution without the extra variable, I am wondering to transform the do...while iteration in while(readline()) but I don't want to mess up the code too much.

tehKaiN:
ah, yes, of course, because getline() triggers eof on stream after reading last line and parsing colors for the last time doesn't occur since the loop condition is met. That loop has kinda strange shape because I wanted to do color parsing in only one place, and szLine already may contain a color definition after the earlier loop for reading comments. So, we can try to remove comment skip loop and add bit more code as below:

``` cpp
 	do {
		std::getline(Source, szLine);
		if(
			szLine != "" &&
			!beginsWith(szLine, "GIMP Palette") && !beginsWith(szLine, "Name:") &&
			!beginsWith(szLine, "Columns:") && !beginsWith(szLine, "#")
		) {
			std::stringstream ss(szLine);
			int r, g, b;
			ss >> r;
			ss >> g;
			ss >> b;
			Color = tRgb(r, g, b);
			Palette.m_vColors.push_back(Color);
		}
	} while(!Source.eof()); 
```

As I said, your original code is fine, so if you don't want to bear with me in finding an elegant solution, just change the variable name and I'll merge it. ;)

tehKaiN:
so, are we ready to merge?

Ozzyboshi:
i think so, i cannot find a more elegant way to to this for now

PR #102, opened 2019-05-28
new functions bitmapLoadFromMem and paletteLoadFromMem implemented

Ozzyboshi:
As the commit message states I implemented bitmapLoadFromMem and paletteLoadFromMem.
This functions should perform the same exact tasks of their LoadFromFile counterparts.
Sometimes I like to store raw data (images and palettes in this case) on variables rather than relying on .plt and .bm files on filesystem loaded at runtime because in the future I am planning to store the whole game on a NDOS disk where I don't have a filesystem, so this features could be handy.

### DIFF
--- a/include/ace/utils/bitmap.h
+++ b/include/ace/utils/bitmap.h
@@ -98,6 +98,24 @@ void bitmapLoadFromFile(
 );
 
 /**
+ *  @brief Loads bitmap data from memory to already existing bitmap.
+ *  If source is smaller than destination, you can use uwStartX & uwStartY
+ *  params to load bitmap on given coords.
+ *
+ *  @param pBitMap    Pointer to destination bitmap
+ *  @param pData      Source bitmap pointer.
+ *  @param uwStartX   Start X-coordinate on destination bitmap, 8-pixel aligned.
+ *  @param uwStartY   Start Y-coordinate on destination bitmap
+ *
+ *  @see bitmapCreate
+ *  @see bitmapCreateFromFile
+ */
+void bitmapLoadFromMem(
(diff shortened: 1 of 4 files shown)


include/ace/utils/bitmap.h
```diff
@@ -97,6 +97,22 @@ void bitmapLoadFromFile(
...
+ *
+ *  @param pBitMap    Pointer to destination bitmap
+ *  @param szFilePath Source bitmap pointer.
+ *  @param uwStartX   Start X-coordinate on destination bitmap, 8-pixel aligned.
+ *  @param uwStartY   Start Y-coordinate on destination bitmap
+ *
+ *  @see bitmapCreate
+ *  @see bitmapCreateFromFile
+ */
+void bitmapLoadFromMem( tBitMap *,const UBYTE*, UWORD , UWORD );
```
tehKaiN: Please add arg names here. It helps in those editors which show function prototype as you type your code - by looking at arg names you don't need to read doxycomment to know what's what.

include/ace/utils/palette.h
```diff
@@ -30,6 +30,15 @@ extern "C" {
  */
 void paletteLoad(char *szFileName, UWORD *pPalette, UBYTE ubMaxLength);
 
+/**
+ *  @brief Loads palette from supplied .plt stored in memory to given address.
+ *  @param szFileName  Palette source pointer.
+ *  @param pPalette    Palette destination pointer.
+ *  @param ubMaxLength Maximum number of colors in palette.
+ */
+void paletteLoadFromMem(const UBYTE* , UWORD *, UBYTE );
```
tehKaiN: same here

tehKaiN:
yep, it's good ;)

PR #103, opened 2019-05-30
Vector base register fixed for 68010+ systems

### LINKED ISSUE #78: Read VBR while taking over OS
I've found such code [here](http://eab.abime.net/showthread.php?t=65430&page=3)

``` c
static APTR GetVBR(void)
{
	APTR vbr = NULL;
	UWORD getvbr[] = {0x4e7a, 0x0801, 0x4e73}; // MOVEC.L VBR,D0 RTE

	if (SysBase->AttnFlags & AFF_68010)
		vbr = (APTR)Supervisor((void *)getvbr);

	return vbr;
}
```

Ozzyboshi:
With this commit interrupts are working on  68010+ systems which don't have VBR at zero.
Code was taken from issue #78 and tested on vampire v600 + coffin r52.
This commit closes #78 hopefully.

### DIFF
--- a/src/ace/managers/system.c
+++ b/src/ace/managers/system.c
@@ -55,7 +55,7 @@ static const tHwIntVector s_pAceHwInterrupts[SYSTEM_INT_VECTOR_COUNT] = {
 	int5Handler, int6Handler, int7Handler
 };
 static volatile tHwIntVector s_pOsHwInterrupts[SYSTEM_INT_VECTOR_COUNT] = {0};
-static volatile tHwIntVector * const s_pHwVectors = 0;
+static volatile tHwIntVector * s_pHwVectors = 0;
 static volatile tAceInterrupt s_pAceInterrupts[SYSTEM_INT_HANDLER_COUNT] = {{0}};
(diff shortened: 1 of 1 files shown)


src/ace/managers/system.c
```diff
@@ -300,7 +302,11 @@ void systemCreate(void) {
 	WaitTOF(); // Wait for interlaced screen to finish
 
 	// get VBR location on 68010+ machine
-	// TODO VBR
+	if (SysBase->AttnFlags & AFF_68010)
+	{
```
tehKaiN: Brace is not codestyle compliant - move it past `if` condition bracket in same line

src/ace/managers/system.c
```diff
@@ -300,7 +302,11 @@ void systemCreate(void) {
...
+		UWORD getvbr[] = {0x4e7a, 0x0801, 0x4e73};
```
tehKaiN: change `getvbr` name to something codestyle compliant - p since it's a pointer (array but it's virtually the same), so perhaps something like `pGetVbrCode`?

PR #105, opened 2019-07-18
Better font loading

Ozzyboshi:
In this PL I fixed a problem about the length of font data loaded into memory, furthermore I added a loadfrommem variant of FontCreate in case the user doesnt want to load the font data from file.
Hope I respected all your coding standard if not let me know.


tehKaiN:
Missing spaces here and there, but yeah, it's good. Thanks!

PR #106, opened 2019-08-06
New inline functions for radians<-->degrees conv

Ozzyboshi:
Added 2 new inline functions for degrees 2 radians and radians 2 degrees conversion.
They work with fix16_t types.


tehKaiN:
rad2deg has parameter name `fAngleDegrees` instead of `fAngleRads` or something. It's important for args to have meaningful name because when IntelliSense suggests function for you, you can determine what to put in given arg just by looking at its name. :)

Rest looks good. While you're at it, plz add a space after each comma, as in `arg1, arg2` and not `arg1,arg2` like it's right now.

tehKaiN:
thanks for fixes!

PR #107, opened 2019-08-16
added paletteColorDim function

Ozzyboshi:
Dimming a single color can be very useful, in my case more than paletteDim.

### DIFF
--- a/include/ace/utils/palette.h
+++ b/include/ace/utils/palette.h
@@ -49,13 +49,11 @@ void paletteLoadFromMem(const UBYTE *pData , UWORD *pPalette, UBYTE ubMaxLength)
 void paletteDim(UWORD *pSource, UWORD *pDest, UBYTE ubColorCount, UBYTE ubLevel);
 
 /**
- *  @brief Dims a single color from the palette palette to given brightness level.
- *  @param pSource      Pointer to source palette.
- *  @param pDest        Pointer to destination palette. May be same as pSource.
- *  @param ubColorIndex Index of the color (from 0 to 31) in palette.
+ *  @brief Dims a single input color to given brightness level.
+ *  @param uwFullColor Full color used as a base to calculate percentage.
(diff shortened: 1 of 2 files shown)


tehKaiN:
I think that having this fn as `UWORD paletteDimColor(UWORD uwColor, ubLevel)` wouldn't be better: you pass just reference color and brightness level, you get modified color in return.

Also, paletteDim should use this fn in a loop, I think.

Ozzyboshi:
something like paletteColorDim( &s_pVp->pPalette[ubColIndex], (15 * fadeOut) / 50);

No more source and destination as parameters?
Source and destination are always the same thing with this approach.
What about a new inline function that calls paletteColorDim? Something like paletteColorChange( &s_pVp->pPalette[ubColIndex], (15 * fadeOut) / 50);
and inside just a regular call to paletteColorDim?

tehKaiN:
I thought something like `uwNewColor = paletteColorDim(uwFullColor, ubFadeRatio)`

Ozzyboshi:
I made a little change that shoud reflect what you asked.
This is a little example how to use it

UWORD s_pPaletteRef[32];
paletteLoadFromMem(PRES1_PALETTE_data, s_pPaletteRef, 1 << s_pVp->ubBPP);
UWORD uwFullColor = s_pPaletteRef[COLOR_INDEX];
UBYTE ubFadeRatio = (15 * fadeIn) / 50;
s_pVp->pPalette[TEXT_COLOR_INDEX] = paletteColorDim(uwFullColor, ubFadeRatio);

what do you think? is this what you planned to dim a single color?

tehKaiN:
Yes, that's exactly it! Now I think you just need to refactor paletteDim() so that it will use paletteColorDim() so that we won't have too much code to maintain. ;)

tehKaiN:
Okay, we're good to go!

PR #108, opened 2019-09-04
Indentation issue in latest commits + wrong call

Ozzyboshi:
PaletteColorDim call was wrong, indentation was not the same of the other functions in the same file, tried to fix it.

### DIFF
--- a/src/ace/utils/palette.c
+++ b/src/ace/utils/palette.c
@@ -40,25 +40,25 @@ void paletteLoadFromMem(const UBYTE* pData, UWORD *pPalette, UBYTE ubMaxLength) 
 }
 
 void paletteDim(UWORD *pSource, UWORD *pDest, UBYTE ubColorCount, UBYTE ubLevel) {
-  for(UBYTE c = 0; c != ubColorCount; ++c) {
-    pDest[c] = paletteGetColorDim(pSource[c],  ubLevel) ;
-  }
+	for(UBYTE c = 0; c != ubColorCount; ++c) {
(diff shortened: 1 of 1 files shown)


Ozzyboshi:
Please take a look at your commit fc0576911d60918ed42d643f750d04b80fbe1735 , should be double the memcpy? In my opinion yes

src/ace/utils/palette.c
```diff
@@ -34,31 +34,31 @@ void paletteLoadFromMem(const UBYTE* pData, UWORD *pPalette, UBYTE ubMaxLength)
 	);
 
 	UBYTE ubPaletteLength = pData[0];
-	memcpy(pPalette, &pData[1], MIN(ubMaxLength, sizeof(UWORD) * ubPaletteLength));
+	memcpy(pPalette, &pData[1], MIN( sizeof(UWORD) * ubMaxLength, sizeof(UWORD) * ubPaletteLength));
```
tehKaiN: you can put sizeof(UWORD) before MIN() so you would have: 

```
sizeof(UWORD) * MIN(ubMaxLength, ubPaletteLength)
```

tehKaiN: Please change in paletteLoad() fn for consistency, from:

```
	if(ubPaletteLength > ubMaxLength) {
		ubPaletteLength = ubMaxLength;
	}
	fileRead(pFile, pPalette, sizeof(UWORD) * ubPaletteLength);
```

to:

```
	fileRead(pFile, pPalette, sizeof(UWORD) * MIN(ubPaletteLength, ubMaxLength));
```

PR #122, opened 2020-06-15
Game manager rewrite

### LINKED ISSUE #33: Game manager tweaks
1. Remove `pView` field and `viewLoad()` stuff
    1. This code doesn't work anyway since gamestate struct's view is set to 0
    2. There may be gamestates which don't have their own view - such as pause menu displayed in border on game's main vport.
2. `gameDestroy()` should call `gamePopState()` in a loop
    1. Less code to maintain
    2. Speed isn't crucial during `gameDestroy()`
3. Remove fadeout remnants from code
    1. This should be shifted to gamestate implementation or viewport managers since in multi-viewport view with color-changing copperlist doing palette fade isn't trivial task.

Since @approxit is responsible for game manager, I'd like to see his opinion in the comments.

approxit:
Old game manager was horrible and way too simple with current needs.

I've removed global game manager instance in favor of manual instancing, which means that game managers can be nested, or even not used in some games. As previous version was managing memory automatically, system was flipped every time, which could be far from optimal. Game manager now works with manually and externally allocated states. In addition two new callbacks was introduced: `Suspend` and `Resume`, as there are cases to run some code before new state is pushed and after some state was popped from game manager.

I'm thinking about renaming `gameState` prefix into just `state`, as it can be more accurate. Also we have first time a manager that is not a singletone.

Closes #33?


include/ace/generic/main.h
```diff
@@ -50,14 +50,12 @@ int main(void) {
 	blitManagerCreate();
 	copCreate();
 
-	gameCreate();
```
tehKaiN, line 53: Since you're removing them from generic `main()`, this is a breaking change. I'm fine with that, but sample code in docs needs updating to reflect those changes. Or, perhaps the generic main should create the first instance of game manager, since most of the games will use it anyway.

include/ace/managers/game.h
```diff
@@ -26,40 +24,60 @@ typedef struct _tGameState {
 	tGameCb cbCreate;
 	tGameCb cbLoop;
 	tGameCb cbDestroy;
-	struct _tGameState *pPrev;
+	tGameCb cbSuspend;
+	tGameCb cbResume;
+	struct _tGameState *pStatePrev;
```
tehKaiN, line 29: It doesn't need to be named `pStatePrev`, since "state" component is deductible from field type and from context - what else could it be if not pointer to the state? It can stay, but I prefer as short names as possible if doesn't obfuscate the meaning.

include/ace/managers/game.h
```diff
@@ -26,40 +24,60 @@ typedef struct _tGameState {
...
 } tGameState;
 
 typedef struct _tGameManager {
-	UBYTE ubStateCount;
-	UBYTE isRunning;
-	tGameState *pStateFirst;
+	tGameState *pState;
```
tehKaiN, line 33: Same here. I think `pFirst` would be better but it's up to you.

include/ace/managers/game.h
```diff
@@ -26,40 +24,60 @@ typedef struct _tGameState {
...
-void gameProcess(void);
-
-void gameClose(void);
-
-void gameKill(char *szError);
+tGameManager *gameManagerCreate(void);
+
+void gameManagerDestroy(
+	tGameManager *pGameManager
+);
```
tehKaiN, line 46: wtf with this formatting? it fits in single line, why each param in one line? makes the header way too big.

src/ace/managers/game.c
```diff
@@ -5,108 +5,128 @@
 #include <ace/managers/game.h>
 #include <ace/managers/system.h> // systemKill
 
-/* Globals */
-tGameManager g_sGameManager;
+UBYTE s_isGameRunning = 1;
```
tehKaiN, line 8: add `static` ffs

src/ace/managers/game.c
```diff
@@ -5,108 +5,128 @@
...
+	pGameManager->pState = pGameState;
+
+	if (pGameManager->pState && pGameManager->pState->cbCreate) {
+		pGameManager->pState->cbCreate();
+	}
+
+	logBlockEnd("gameStateChange()");
 }
+
+void gameStateProcess(tGameManager *pGameManager) {
```
tehKaiN, line 128: add `const` to arg, I think

src/ace/managers/game.c
```diff
@@ -5,108 +5,128 @@
...
+		pGameManager->pState = pGameManager->pState->pStatePrev;
 	}
+
+	logBlockEnd("gameStatePopAll()");
 }
 
-void gameKill(char *szError) {
-	gameDestroy();
-	systemKill(szError);
+void gameStateChange(tGameManager *pGameManager, tGameState *pGameState) {
```
tehKaiN, line 112: - change to `const tGameState *pGameState`, I think
- what happens if pGameState is null?

src/ace/managers/game.c
```diff
@@ -5,108 +5,128 @@
 #include <ace/managers/game.h>
 #include <ace/managers/system.h> // systemKill
```
tehKaiN, line 6: ace/managers/system.h is no loger used, I think

include/ace/managers/game.h
```diff
@@ -11,8 +11,6 @@ extern "C" {
 
 #include <ace/types.h> // Amiga typedefs
 
-#include <ace/types.h>
-
 #include <ace/managers/timer.h>
 #include <ace/managers/key.h>
```
tehKaiN, line 15: are those includes used?

src/ace/managers/game.c
```diff
@@ -5,108 +5,128 @@
...
-	// I guess it should stay here, makes each push a bit faster
-	g_sGameManager.isRunning = 1;
-	logBlockEnd("gameCreate()");
+void gameExit(void) {
+	s_isGameRunning = 0;
 }
 
 UBYTE gameIsRunning(void) {
-	return g_sGameManager.isRunning;
+	return s_isGameRunning;
```
tehKaiN, line 16: I have a problem with this variable.

- it doesn't get set on gamestate manager creation
- it doesn't get set/unset when you push/pop states so your game must call gameExit at some point.

So if you're just pushing and popping your states, gameIsRunning doesn't change to 0 by itself so the game won't close automatically if you just check if game is running like in the old way.

src/ace/managers/game.c
approxit: At first, I had `const`s, but unfortunately compiler was complaining a lot without type casting.

src/ace/managers/game.c
approxit: At first, I had consts, but unfortunately compiler was complaining a lot without type casting.

Everything will blow up. I'd like to use some kind of assert to handle that kind of errors in debug builds.

include/ace/generic/main.h
approxit: I've created generic global state manager.

src/ace/managers/game.c
tehKaiN:
```
#if defined(ACE_DEBUG)
  bugCheck();
#endif
```

include/ace/managers/game.h
approxit: I'll stick with `pState`.

src/ace/managers/game.c
approxit: Variable was tossed to garbage bin.

include/ace/managers/game.h
tehKaiN: wouldn't `pCurrent` be better?

src/ace/managers/game.c
approxit: As discussed offline, variable is not actually needed, as same kind of logic could be archived by checking if current state is not null.

approxit:
Also, I've renamed `game` to `state`.

include/ace/managers/game.h
approxit: True, It will be more aligned with `pPrev`.

src/ace/managers/state.c
```diff
@@ -0,0 +1,136 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <ace/managers/state.h>
+
+#include <ace/managers/log.h>
+
+/* Functions */
+tStateManager *stateManagerCreate() {
```
tehKaiN, line 10: in C, empty parameter list means "0 or 1 arguments", whereas in C++ it means "0 arguments". It' best to write `void` to strictly tell compiler that you mean 0 args.

src/ace/managers/state.c
```diff
@@ -0,0 +1,136 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <ace/managers/state.h>
+
+#include <ace/managers/log.h>
```
tehKaiN, line 7: no spaces between includes - don't bloat file with empty lines

src/ace/managers/state.c
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	logBlockEnd("stateChange()");
+}
+
+void stateProcess(tStateManager *pStateManager) {
+	if (pStateManager->pCurrent && pStateManager->pCurrent->cbLoop) {
+		pStateManager->pCurrent->cbLoop();
+	}
+}
```
tehKaiN, line 136: configure vscode to put newline at the end of file. It's common practice in C world, it may produce compiler warnings in other version. Also, IIRC vbcc had problem with lack of newlines. Not that we support it, but it's a good practice. Even GH shows fancy icon that you've ommited it. Almost all of POSIX world uses such convention. https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

include/ace/managers/state.h
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Cleans up after state manager.
+ * @param pStateManager: Pointer to state manager previously created 
+ *        with stateManagerCreate.
+ * @see stateManagerCreate()
+ */
+void stateManagerDestroy(tStateManager *pStateManager);
+
+
```
tehKaiN, line 47: no double newlines plz

include/ace/managers/state.h
```diff
@@ -0,0 +1,127 @@
...
+extern "C" {
+#endif
+
+#include <ace/types.h> // Amiga typedefs
+
+/* Types */
+
+typedef void (*tStateCb)(void);
+
+typedef struct _tState {
```
tehKaiN, line 18: You should document which struct fields are mandatory when creating state struct by hand, and all special rules. For instance, should pPrev be set to zero? Or is the value irrelevant and will be reset by push()?

EDIT: now I understand that you should use stateCreate(). This should be pointed in doxy for struct.

include/ace/generic/main.h
```diff
@@ -41,6 +41,8 @@ void __stack_chk_fail(void) {
 }
 #endif
 
+tStateManager *g_pGameStateManager;
```
tehKaiN, line 44: I have a problem with this and don't know how to solve it.

- Suppose someone wants to create files for state A and state B
- main.c includes generic/main.h and pushes state to A
- a.c  includes generic/main.h and pushes state to B
- then you have redeclaration of g_pGameStateManager

One has to create header extern for that variable. This is not documented, nor intuitive for beginners. So either it should be very clearly documented in docs *and* main.h or done in some other manner - dunno how though.

include/ace/managers/state.h
```diff
@@ -0,0 +1,127 @@
...
+ * @param pPrev: Pointer to previous state. Zero if there is no previous state.
+ * @see stateDestroy()
+ */
+tState *stateCreate(
+	tStateCb cbCreate,
+	tStateCb cbLoop,
+	tStateCb cbDestroy,
+	tStateCb cbSuspend,
+	tStateCb cbResume,
+	tState *pPrev
```
tehKaiN, line 64: wtf is with this formatting? Don't use one arg per line.

include/ace/generic/main.h
approxit: That's why I removed default state manager... And I thing that is more correct way in this case.

src/ace/managers/game.c
approxit: Actually logging is doing the job - it prints all pointer values. If something would crash - last entry is showing zero value.

tehKaiN: yeah but there's a convenience in being able to search logs for `ERR:` and find exactly what you're doing wrong. ;)

include/ace/generic/main.h
tehKaiN: Yes and no. Yes because it solves multiple definition thingy, no because generic main was supposed to remove boilerplate used in 95% of cases, and gamestate manager is part of it.

I don't have a better idea how to solve it so yeah, you can remove it. Be sure to fix the docs/tutorial though. Also, while you're at it, refactor showcase code to runnable state.

tehKaiN:
Todo:

- update docs
- make showcase work again

approxit:
I've updated docs and showcase a bit.

Unfortunately showcase bugs out if we run it second time. Any ideas?

src/ace/managers/state.c
```diff
@@ -0,0 +1,181 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <ace/managers/state.h>
+#include <ace/managers/log.h>
+
+/* Functions */
+
+#ifdef ACE_DEBUG
+void _checkPointers(const char *szPointerName, void *pPointer) {
```
tehKaiN, line 11: - you're not checking pointer**s** here
- name could be more meaningful, like `checkNull()`.
- fn could also take file/line args and be called via macro - see `blitCheck()`.
- I think you could also automatically generate pointer name as an arg: https://stackoverflow.com/questions/7561828/variable-name-from-macro-argument
- it could be declared as `static inline` and moved to ace/macros.h or something.

src/ace/managers/state.c
```diff
@@ -0,0 +1,181 @@
...
+	if (!pPointer) {
+		logWrite("ERR: Pointer %s is zero! Crash is eminent!\n", szPointerName);
```
tehKaiN, line 13:
```suggestion
		logWrite("ERR: Pointer %s is zero! Crash imminent!\n", szPointerName);
```

src/ace/managers/state.c
```diff
@@ -0,0 +1,181 @@
...
+
+	logBlockEnd("stateManagerCreate()");
+
+	return pStateManager;
+}
+
+void stateManagerDestroy(tStateManager *pStateManager) {
+	logBlockBegin("stateManagerDestroy(pStateManager: %p)", pStateManager);
+
+#ifdef ACE_DEBUG
```
tehKaiN, line 31: if you call fn via macro, the macro def can be empty in release builds, so no need for ifdefs butchering our code.

docs/programming/ace_in_a_nutshell.md
```diff
@@ -141,8 +145,8 @@ OS  --CREATE-->  menuCommon  --PUSH-->  menuMain
 - lastly, _menuCommon_ should in its loop examine why its loop got called and
   decide what to do next. It can be done by e.g. examining some global variable.
   If it was because menuMapSelect requested launching game, you change state
-  to _game_, otherwise you call `gameClose` or `gamePopState` which will
-  effectively close the game.
+  to _game_, otherwise you call `gameExit` which will effectively close the
+  game.
```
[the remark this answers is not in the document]
approxit: It will be better to move it to separate issue.

showcase/src/game.c
```diff
@@ -0,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
approxit: Well... As we're using generic main, I tried to keep names separate as "framework runs you".

src/ace/managers/state.c
approxit: Done, but this code can't be in `macros.h` in this shape - logs causing circular includes.

docs/programming/ace_in_a_nutshell.md
```diff
@@ -159,6 +163,19 @@ Instead of writing `main()` function you just `#include` this file and define:
 - `genericLoop()` - called in a loop until game gets closed
 - `genericDestroy()` - for freeing previously created managers
 
+If you preffer exiting your game in other ways than by calling `gameExit`, you
```
tehKaiN, line 166:
```suggestion
If you prefer exiting your game in other ways than by calling `gameExit`, you
```

docs/programming/ace_in_a_nutshell.md
```diff
@@ -159,6 +163,19 @@ Instead of writing `main()` function you just `#include` this file and define:
...
+can tune game main loop condition by defining `GENERIC_MAIN_LOOP_CONDITION`
```
tehKaiN, line 167: And what is generic main.h? It's not mentioned until this point. This chapter is about basic game state concept, not about generic main.

Maybe move it to tutorial or describe it in generic.h as comment?

docs/programming/ace_in_a_nutshell.md
```diff
@@ -159,6 +163,19 @@ Instead of writing `main()` function you just `#include` this file and define:
...
+before `#include`. For e.g.:
```
tehKaiN, line 168: Use either 'For example', or 'E.g.' (exempli gratia)

docs/programming/ace_in_a_nutshell.md
```diff
@@ -159,6 +163,19 @@ Instead of writing `main()` function you just `#include` this file and define:
...
+
+``` c
+#define GENERIC_MAIN_LOOP_CONDITION g_pGameStateManager->pCurrent
+#include <ace/generic/main.h>
+```
+
+At this point we changed generic main loop to check if there is any state in
```
tehKaiN, line 175:
```suggestion
This changes generic main loop's behavior to check if there is any state in 
```

docs/programming/ace_in_a_nutshell.md
approxit: This chapter as a whole literally explains everything about `main.h`... So... Where is the problem?

tehKaiN: sorry, I was basing this on my memory and diffs, haven't looked at the whole text. You're right, it's good.

SamuraiCrow:
Is "prefer" spelled correctly in the current draft?

approxit:
By having ability to customize generic loop condition you can have both: `isGameRunning() && g_pGameStateManager->pCurrent`. After calling `gameExit()` or poping out all states, loop is chuting down, and `genericDestroy` happens. When state manager is allocated by `stateManagerCreate` you need call its destroy, which happens in `genericDestroy()`. I'll add this case in showcase.

tehKaiN:
> 
> 
> Is "prefer" spelled correctly in the current draft?

google says yes

PR #126, opened 2020-07-25
Generic main documentation

tehKaiN:
I've documented some stuff in ACE based on frequent questions I get.

### DIFF
--- a/docs/programming/os.md
+++ b/docs/programming/os.md
@@ -1,0 +1,87 @@
+# Working with and without OS
+
+To squeeze as much performance as possible, ACE allows for disabling Amiga OS and re-enabling it when needed.
+Typically, 95% of stuff which game performs doesn't need operating system, so you're safe to have it disabled when your game runs.
+Things that require having OS enabled:
+
+- memory allocation and deallocation - most `ace/managers/memory.h` functions
+- logging to file - all `ace/managers/log.h` functions if log file is created and when in debug mode
+- general file access - all `ace/utils/file.h` and `ace/utils/dir.h` functions
+- system .library and .device usage
+- multithreading, communicating with other apps
+
+## Manually managing OS state
+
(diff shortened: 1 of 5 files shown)


docs/programming/os.md
```diff
@@ -0,0 +1,87 @@
...
+s_pBm1 = bitmapCreate(320, 256, 5);         // They have systemUse()/systemUnuse calls
+s_pBm2 = fontCreate("test.fnt");            // But they won't do anything since OS
+logWrite("I'm done initializing\n");        // Is already enabled
+systemUnuse(); // Disable OS once when we're done with it
+```
+
+## Best practices
+
+Try to have OS disabled at the end of gamestate creation, and to re-enable it at the beginning of gamestate destruction.
+This way, the game loop performs with maximum performance, whereas you have OS at your disposal in init/deinit phases.
```
approxit, line 48:
```suggestion
This way, the game loop performs with maximum performance, whereas you have OS at your disposal in create/destroy phases.
```
It's better to keep existings terms instead of providing new ones in docs.

docs/programming/os.md
```diff
@@ -0,0 +1,87 @@
...
+
+Be careful when using state pushing/popping.
+If pushed state needs OS, be sure to have it re-enabled at the beginning of gamestate create phase and disabled at the end of gamestate destroy phase.
+You can use onSuspend and onResume callbacks for that.
```
approxit, line 52: I think that this proposal to handle OS while pushing / popping is way worse approach than proposal from next paragraph. Also OS handling in create/destroy is way more popular than in push/pop, so it should be mentioned first in order. Push/pop should be mentioned as if game logic forces you to use other methods than create/destroy.

tehKaiN: Yeah, you're right. I've removed it and changed order so that push/pop warning is after the grouping advice.

PR #134, opened 2021-03-07
Print names of cop registers in debug build logs

timfel:
I found this useful. It prints SKIP rather than WAIT if the command has the skip bit set, and it prints the names for known register target addresses in MOVE instructions


src/ace/managers/copper.c
```diff
@@ -57,19 +57,285 @@ void copSwapBuffers(void) {
...
+    "VSSTRT",
+    "HCENTER",
+    "DIWHIGH",
+    NULL
+};
+#endif
+
 void copDumpCmd(tCopCmd *pCmd) {
 	if(pCmd->sWait.bfIsWait) {
+		const char* cmd = pCmd->sWait.bfIsSkip ? "SKIP" : "WAIT";
```
tehKaiN, line 311: please use prefix notation. For zero-terminated strings, it's `sz`, so should be `szCmd`

src/ace/managers/copper.c
```diff
@@ -57,19 +57,285 @@ void copSwapBuffers(void) {
 	pCopList->pBackBfr = pTmp;
 }
 
+#ifdef ACE_DEBUG
+static const char* _custom_register_names[256] = {
```
tehKaiN, line 61: we use CamelCaps for names, preceded with prefix.

- Since it's an array, it is a pointer, hence the `p` prefix.
- Since it's the `static` global var, additional `s_` prefix is added.

So, the final name should be `s_pCustomRegNames` or something.

Also, since this array is used only in `copDumpCmd()`, perhaps add it as `static const` var inside fn? If so, skip the `s_` prefix.

src/ace/managers/copper.c
```diff
@@ -57,19 +57,285 @@ void copSwapBuffers(void) {
...
-			"@%p: %08lX - WAIT: %hu,%hu%s\n",
-			pCmd, pCmd->ulCode, pCmd->sWait.bfWaitX << 1, pCmd->sWait.bfWaitY,
+			"@%p: %08lX - %s: %hu,%hu%s\n",
+			pCmd, pCmd->ulCode, cmd, pCmd->sWait.bfWaitX << 1, pCmd->sWait.bfWaitY,
 			!pCmd->sWait.bfBlitterIgnore ? " & blit done" : ""
 		);
 	}
 	else {
+#ifdef ACE_DEBUG
+		const char* name = _custom_register_names[pCmd->sMove.bfDestAddr >> 1];
```
tehKaiN, line 320: This var needs name changing too - `szRegName` perhaps?

tehKaiN (changes requested):
Looks good, but needs some variable name to be consistent with rest of the code.

https://github.com/AmigaPorts/ACE/blob/master/docs/contributing/codestyle.md#naming-conventions

PR #135, opened 2021-03-07
Fix logfile writing

timfel:
I noticed this hangs on Linux when the logfile is opened for writing if the calls are in the previous order.

### DIFF
--- a/include/ace/generic/main.h
+++ b/include/ace/generic/main.h
@@ -50,8 +50,8 @@ void __stack_chk_fail(void) {
 #endif
 
 int main(void) {
-	logOpen(GENERIC_MAIN_LOG_PATH);
 	systemCreate();
+	logOpen(GENERIC_MAIN_LOG_PATH);
 	memCreate();
 	timerCreate();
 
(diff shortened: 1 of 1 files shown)


tehKaiN:
Uhhh, it's not that simple. The idea is to have some kind of logging to have output from systemCreate. Since file logging may and will crash before OS is initialized, it should use UAE console, if ACE is configured to use UAE console logging. This should be handled by log manager - if log file handle is null, it will use only console, if available. If log file handle is non-null, it should use both console and the file.

It's strange that on Linux this gives a problem. Can you give me more details on it?

timfel:
It'll create the logfile and then hang forever. I tried checking with the FS-UAE console debugger, but wasn't sure what I was looking for

tehKaiN:
Aw crap, I've got the pull request backwards! I was certain that currently the log file gets created after `systemCreate()`. Indeed, this pull request is valid and I completely missed the erratic behavior because I use NULL log file name, so it's skipped and only console is used for logging.

Sorry for confusion and thanks for fixing!

PR #136, opened 2021-03-14
RAW coplist support in scrollbuffer and tilebuffer

timfel:
It's not finished, at the very least it's missing a showcase that uses it. But I wanted to open it to get feedback already.

### DIFF
--- a/include/ace/managers/viewport/tilebuffer.h
+++ b/include/ace/managers/viewport/tilebuffer.h
@@ -170,11 +170,6 @@ void tileBufferDestroy(tTileBufferManager *pManager);
  */
 void tileBufferProcess(tTileBufferManager *pManager);
 
-void tileBufferReset(
-	tTileBufferManager *pManager, UWORD uwTileX, UWORD uwTileY,
-	UBYTE ubBitmapFlags, UBYTE isDblBuf
-);
-
 /**
  * Redraws tiles on whole screen.
  * Use for init or something like that, as it's slooooooooow.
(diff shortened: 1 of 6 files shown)


src/ace/managers/viewport/scrollbuffer.c
```diff
@@ -75,22 +75,42 @@ tScrollBufferManager *scrollBufferCreate(void *pTags, ...) {
 	// Create copperlist entries
 	pCopList = pVPort->pView->pCopList;
 	if(pCopList->ubMode == COPPER_MODE_BLOCK) {
-		pManager->pStartBlock = copBlockCreate(
+		pManager->copStart.pStartBlock = copBlockCreate(
```
tehKaiN, line 78: those can be anonymous unions to skip the copStart/copBreak stuff, or if not, union contents can have shorter names to not duplicate the "start/break" and "copper" info from union name.

Something like `pManager->copStart.pBlock` and `pManager->copStart.uwOffset`

include/ace/managers/viewport/tilebuffer.h
```diff
@@ -170,11 +170,6 @@ void tileBufferDestroy(tTileBufferManager *pManager);
  */
 void tileBufferProcess(tTileBufferManager *pManager);
 
-void tileBufferReset(
-	tTileBufferManager *pManager, UWORD uwTileX, UWORD uwTileY,
-	UBYTE ubBitmapFlags, UBYTE isDblBuf
-);
-
```
tehKaiN, line 177: why removed tileBufferReset fn?

tehKaiN (changes requested):
Looks good but I'm a bit confused why you've removed tileBufferReset() from tilebuffer.h header - this is completely valid non-static fn to call when e.g. reloading levels.

Or am I'm missing something due to github diff view? ;)

include/ace/managers/viewport/tilebuffer.h
timfel: you're right, I moved it to `tilebuffer.c` for no good reason.

src/ace/managers/viewport/scrollbuffer.c
timfel: Made it anonymous

include/ace/managers/viewport/tilebuffer.h
timfel: moved back to header

tehKaiN (approved):

PR #143, opened 2021-10-02
Ptplayer c

### LINKED ISSUE #26: Joy manager init should accept parameter on which ports it's listening
Not every game should listen for additional joys on parallel port - there could be other device plugged in.

### LINKED ISSUE #97: Add MOD replayer
Created as a continuation of #32 , see comments there.

I'll try to rewrite ptplayer into C so that it can be easily maintained by me and integrated into ACE across all the code refactoring we are yet to experience ;)

- [x] rewrite ptplayer to C
- [x] add OS enabling/disabling support
- [x] fix bugs on different CPU speeds
- [x] debug basic protracker commands
- [x] remove handling of audio interrupts in ACE - also when OS is enabled

tehKaiN:
Here we go - 200 commits, 200 file changes, a tiny little pull request. Added protracker player based on asm version of ptplayer, also fixed lots of bugs and done minor improvements here and there because of scope creep.

This works relatively stable in GermZ game, so I'm going to merge it by tuesday. If anyone wants to review and point stuff by that time, feel free to do so!

closes #97 
closes #26


src/ace/managers/memory.c
```diff
@@ -23,14 +25,20 @@ typedef struct _tMemEntry {
...
 
 static void _memEntryAdd(
 	void *pAddr, ULONG ulSize, UWORD uwLine, const char *szFile
 ) {
+	// Ozzyboshi discovered that memory allocation works without re-enabling
+	// the OS. After analyzing what's under the hood, it looks like there are only
+	// Forbid/Permit calls inside OS mem fns and the rest is just plain code.
+	// Still, OS mem fns can be patched by Snoopdos and/or newer OS versions,
+	// so I guess the safest approach is not to assume anything and wake OS up
+	// in case it needs other components on some exotic configs.
```
timfel, line 40: I guess we don't care about the performance impact here, since the allocation itself will probably dominate?

src/ace/utils/string.c
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <ace/utils/string.h>
+
+char *stringDecimalFromULong(ULONG ulVal, char *pDst) {
```
timfel, line 7: Formatting looks weirdly rendered on github, tabs vs spaces issue?

src/ace/managers/memory.c
tehKaiN: Allocations are slow by themselves, but there's a problem with user experience of OS wakeup.

The problem is that in order to allocate the memory you need to wake OS up currently, and this implies waiting for at least 2 frames in order to wake up an OS and the same to put it to sleep, and when OS is allive our interrupt handlers aren't working - specifically CIA timers needed for music playback. Now think of the loading between levels where you deallocate things. Skipping the OS wakeup would allow you to play back music in the meantime without glitches, now you either need to pause the playback or fade the volume out to mask the problem.

Still, I think the better way is to add support for interrupts when OS is alive, so things will be slow but working. Also better with the far future of making OS disabling completely optional.

tehKaiN:
No more objections? Let's merge this and watch the world burn!

PR #145, opened 2021-10-07
Windows tools fixes

timfel:
Just a few build fixes to work with the tools on windows


tehKaiN:
I was wondering why I haven't implemented .bm and interleaved export for tileset_conv and I remembered - the idea was to generate intermediate files in PNG format and then convert to .bm using bitmap_conv in the last step. This was mostly to not duplicate the tool functionalities (especially if there were more conversion params in the future), but I guess being able to generate .bm directly is not that bad idea.

PR #158, opened 2022-07-29
Compile tools again with VS2019

timfel:
Just some minor things to make it compile again with msvc 19

### DIFF
--- a/tools/fmt/include/fmt/core.h
+++ b/tools/fmt/include/fmt/core.h
@@ -269,6 +269,7 @@
 
 #ifndef FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 #  if defined(__cpp_nontype_template_args) &&                \
+      (!defined(_MSC_VER) || _MSC_VER >= 1930) &&            \
       ((FMT_GCC_VERSION >= 903 && __cplusplus >= 201709L) || \
        __cpp_nontype_template_args >= 201911L)
 #    define FMT_USE_NONTYPE_TEMPLATE_PARAMETERS 1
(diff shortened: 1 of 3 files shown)


tools/fmt/include/fmt/core.h
```diff
@@ -269,6 +269,7 @@
 
 #ifndef FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 #  if defined(__cpp_nontype_template_args) &&                \
+      (!defined(_MSC_VER) || _MSC_VER >= 1930) &&            \
```
tehKaiN, line 272: huh, that's strange, maybe we should once again update fmt files from main repo

tools/src/common/math.h
```diff
@@ -5,7 +5,11 @@
 #ifndef _ACE_TOOLS_COMMON_MATH_H_
 #define _ACE_TOOLS_COMMON_MATH_H_
 
+#if defined(_MSC_VER) && _MSC_VER < 1930
+auto roundToMultipleOf(uint64_t Value, int64_t Multiple) {
```
tehKaiN, line 9: maybe we could use a template here?

```suggestion
template<typename T>
auto roundToMultipleOf(T Value, T Multiple) {
```

tools/fmt/include/fmt/format.h
```diff
@@ -33,6 +33,7 @@
 #ifndef FMT_FORMAT_H_
 #define FMT_FORMAT_H_
 
+#include <bit>           // std::bit_cast
```
tehKaiN, line 36: funny that quite new GCC versions as well as vc2022 works without it

tehKaiN (approved):

PR #159, opened 2022-10-11
Fixed clearing interupt instead of dmacon when disabling handler

rveilleux:
Fixed INTF_SETCLR (even if same value as DMAF_SETCLR), to make code more accurate
This bug was messing up sprites when ptplayer is set to use PTPLAYER_USE_VBL: since INTB_VERTB is equal to DMAB_SPRITE.


tehKaiN:
Such a silly mistake, how did I miss it? Thanks for spotting it!

PR #161, opened 2022-10-12
Small Ptplayer optim and fix

rveilleux:
Fixed timer A interrupt cleanup code when using VBL
and slight performance boost when loading MOD (especially from floppy) by not query getFileSize() twice


tehKaiN:
nice! I'm currently in the process in updating its version and thought for a while about scrapping this mode. Since I see you really need this I'll take extra care to not break things in this mode.

PR #165, opened 2022-10-17
Ptplayer bugfixes

rveilleux:
A pull request for a variety of fixes to ptplayer:
I tried several MODs and now they all sound fantastic. They are probably still some little bugs on some effects, and I have a least two questions for you (KaiN) for further discussion / investigation for later.

### DIFF
--- a/src/ace/managers/ptplayer.c
+++ b/src/ace/managers/ptplayer.c
@@ -2485,7 +2485,7 @@ static void set_period(
 		// );
 
 		pChannelReg->ac_ptr = pChannelData->n_start;
-		pChannelReg->ac_len = pChannelData->n_length;
+		pChannelReg->ac_len = pChannelData->n_reallength;
 		pChannelReg->ac_per = uwPeriod;
 		mt_dmaon |= pChannelData->uwDmaFlag;
 	}
--- a/src/ace/managers/ptplayer.c
+++ b/src/ace/managers/ptplayer.c
@@ -2517,7 +2517,7 @@ static void set_sampleoffset(
 	else {
 		pChannelData->n_sampleoffset = ubArg;
 	}
-	UWORD uwLength = ubArg << 7;
+	UWORD uwLength = ubArg << 6;
 	if(uwLength < pChannelData->n_length) {
 		pChannelData->n_length -= uwLength;
 		pChannelData->n_start += uwLength;
--- a/src/ace/managers/ptplayer.c
+++ b/src/ace/managers/ptplayer.c
@@ -31,6 +31,26 @@
 
 //---------------------------------------------------------------------- DEFINES
 
+/**
+ * @brief Minimum safe CIA timer ticks count after which Paula channel regs can
+ * be set for next sample.
+ * 
+ * A raster line has a DMA slot for each audio channel. Depending on the current
+ * AUDxPER of the audio channel some of these DMA slots are not used.
+ * For example: when playing a very high frequency the DMA might have to read
+ * a new sample every line. But when playing a very low frequency there might be
+ * 10 or more raster lines without DMA activity for this channel.
+ * 
+ * The problem is that changes like DMA-on/off, or setting a new sample-pointer
+ * in AUDxLC, are only recognized by Paula when the next sample data are read
+ * from Chip-RAM.
+ * 
+ * So we have to be prepared for the worst case, which means that
+ * up to 11 raster lines may pass without DMA-activity (reading a new sample),
+ * when playing the lowest note from the lowest Protracker octave.
+ */
+#define TIMERB_TICKS (576)
+
 // Patterns - each has 64 rows, each row has 4 notes, each note has 4 bytes.
 #define MOD_ROWS_IN_PATTERN 64
 #define MOD_NOTES_PER_ROW 4
@@ -1395,8 +1415,6 @@ static void intSetRep(volatile tCustom *pCustom) {
 	systemSetCiaInt(CIA_B, CIAICRB_TIMER_B, 0, 0);
 }
 
-#define TIMERB_TICKS (576)
-
 // One-shot TimerB interrupt to set repeat samples after another TIMERB_TICKS ticks.
 static void mt_TimerBsetrep(
 	REGARG(volatile tCustom *pCustom, "a0"),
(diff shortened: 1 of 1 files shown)


src/ace/managers/ptplayer.c
```diff
@@ -1215,7 +1216,7 @@ static void mt_playvoice(
 		pChannelData->n_reallength = uwSampleLength;
 
 		// Determine period table from fine-tune parameter
-		UBYTE ubFineTune = pSampleDef->ubFineTune; // d3
+		UBYTE ubFineTune = pSampleDef->ubFineTune & 0xF;
```
tehKaiN, line 1219: This worries me a bit. Is it for some kind of malformed MODs or perhaps there are values above 15 which need some kind of different handling for topmost bits? A comment in the source code would be helpful here.

src/ace/managers/ptplayer.c
```diff
@@ -1231,7 +1232,7 @@ static void mt_playvoice(
 			// Set repeat
 			pChannelData->n_looped = 1;
 			pChannelData->n_replen = pSampleDef->uwRepeatLength;
-			uwSampleLength = pSampleDef->uwRepeatLength + ubFineTune * 2;
+			uwSampleLength = pSampleDef->uwRepeatLength;
```
tehKaiN, line 1235: note to self: check against original .s file if it was there  or if it's some kind of bug due to asm->c translation

src/ace/managers/ptplayer.c
```diff
@@ -1243,7 +1244,7 @@ static void mt_playvoice(
 	}
 
 	// inlined set_regs function here:
-	if(!pVoice->uwNote) {
+	if(!(pVoice->uwNote & 0xFFF)) {
```
tehKaiN, line 1247: note to self: check against .s file

src/ace/managers/ptplayer.c
```diff
@@ -1393,7 +1394,9 @@ static void intSetRep(volatile tCustom *pCustom) {
 	systemSetCiaInt(CIA_B, CIAICRB_TIMER_B, 0, 0);
 }
 
-// One-shot TimerB interrupt to set repeat samples after another 576 ticks.
+#define TIMERB_TICKS (576)
```
tehKaiN, line 1397: this should be on top of file in the DEFINES section

src/ace/managers/ptplayer.c
```diff
@@ -1791,7 +1794,7 @@ static void mt_toneporta_nc(
 	tChannelStatus *pChannelData, volatile tChannelRegs *pChannelReg
 ) {
 	if(pChannelData->n_wantedperiod) {
-		UWORD uwNew;
+		WORD uwNew;
```
tehKaiN, line 1797: since you've changed `UWORD` to `WORD`, the prefix should be changed from `uw` to `w`

src/ace/managers/ptplayer.c
```diff
@@ -2514,7 +2517,7 @@ static void set_sampleoffset(
 	else {
 		pChannelData->n_sampleoffset = ubArg;
 	}
-	UWORD uwLength = ubArg << 7;
+	UWORD uwLength = ubArg << 6;
```
tehKaiN, line 2520: I'd reword `<< 6` to something like `* (256 / sizeof(UWORD))` to indicate the intent - ubArg is length in 256s of bytes, but we need the length in words, I assume. Perhaps also add a comment if that kind of notation is not clear enough.

src/ace/managers/ptplayer.c
```diff
@@ -2533,15 +2536,15 @@ static void set_toneporta(
...
-	}
+	// if(ubPeriodPos) {
+	// 	// One before for less/equal
+	// 	--ubPeriodPos;
+	// }
+
+	// if(pChannelData->n_minusft && ubPeriodPos) {
+	// 	// Negative fine tune? Then take the previous period.
+	// 	--ubPeriodPos;
+	// }
```
tehKaiN, line 2547: what's going on here?

tehKaiN (changes requested):
Overall I'd approve and merge it as-is, but the last commented-out code needs some more comments on why it is this way.

Tomorrow I'll try to check 2 things mentioned in review against original assembly to see if it's my mistake or if there's some kind of intent for current state of code.

rveilleux:
Hi KaIN,

 thank you for the feedback!

I did put some explanations in my individual commits: I wonder if the pullrequest brings them? Else perhaps you can gaze a my fork repo?

Also I've made a potential great speed improvement on findPeriod that we can discuss in another thread if you are interested.

Rem.

tehKaiN:
now I see that commit history describe the changes, but this kind of documenting is problematic - it isn't that easy to inspect without checking out the branch and using some kind of IDE gutter which shows the commit message near each of its related lines. Even then it only shows only message for latest change. 

I strongly prefer to have this type of documentation as code comments. I'd be more than grateful if you could transfer most of the helpful messages to comments.

Re optimizations, I think it's best to keep PRs as concise as possible. Let's merge in this and discuss the optimization separately.

tehKaiN:
I'm gonna take a look on that original .s file and will merge it today. Sorry for delay!

src/ace/managers/ptplayer.c
tehKaiN: from @rveilleux 's commit message:
> this change is not 100% certain because the original ASM code does something similar, but since it does cmp.w (a0)+,d6, this always increment a0 one element past where it should.
> Might need revist, but at least now it sounds accurate and not out-of-tune

src/ace/managers/ptplayer.c
tehKaiN: in mt_playvoice there's this fragment

```m68k
	; set repeat
	move.w	(a0),d0
	move.w	d0,n_replen(a2)
	add.w	d3,d0
	add.l	d3,d2
	add.l	d3,d2
	bra	set_loop
```

And earlier on there's calculation of fine tune

```m68k
	moveq	#0,d3
	move.b	(a0)+,d3 ;;; d3 := nibble
	add.w	d3,d3      ;;; nibble *= 2
```

What I overlooked is that there's also between them:

```
	move.w	(a0)+,d3		; repeat offset
```

src/ace/managers/ptplayer.c
tehKaiN: I've checked all the MOD docs I know and haven't found any info about upper nibble except that it must be zero. I think the best would be having some kind of sanitizer for that value, but that's work for separate issue. A todo comment here should suffice.

```suggestion
		// TODO: sanitize this value to only have lower nibble set when loading MOD/samplepack
		UBYTE ubFineTune = pSampleDef->ubFineTune & 0xF;
```

src/ace/managers/ptplayer.c
tehKaiN: I can't seem to add suggestion for that change, too much lines :(

src/ace/managers/ptplayer.c
tehKaiN:
```suggestion
	// Offset is in 256s of bytes, length is in words
	UWORD uwLength = ubArg * (256 / sizeof(UWORD));
```

src/ace/managers/ptplayer.c
```diff
@@ -2533,15 +2536,15 @@ static void set_toneporta(
...
-	if(ubPeriodPos) {
-		// One before for less/equal
-		--ubPeriodPos;
-	}
-
-	if(pChannelData->n_minusft && ubPeriodPos) {
-		// Negative fine tune? Then take the previous period.
-		--ubPeriodPos;
-	}
+	// if(ubPeriodPos) {
```
tehKaiN, line 2539:
```suggestion
	// Original ASM code does something similar, but without those lines it sounds accurate and not out-of-tune
	// if(ubPeriodPos) {
```

tehKaiN (commented):
I've added suggestions where possible. Apply them, move define to the top with others and we can merge this.

tehKaiN (approved):
Looks like somehow I could commit to reviewed branch. So I did the fixes myself. ;)

PR #172, opened 2022-11-29
fix duplicate samples (all uppercase)

Luc3kLMC:


tools/src/common/mod.cpp
```diff
@@ -8,6 +8,8 @@
 #include <fmt/format.h>
 #include "endian.h"
 
+#define SAMPLE_NAME_SIZE 22
```
tehKaiN, line 11: nice!

tools/src/common/mod.cpp
```diff
@@ -55,6 +57,9 @@ tMod::tMod(const std::string &szFileName)
 
 		// Data read successfully, fill sample info
 		tSample Sample;
+		for (uint8_t i = 0; i < SAMPLE_NAME_SIZE; ++i){
```
tehKaiN, line 60: we already have `i` variable in `for` loop above, it's best to not redeclare variables like that. Perhaps it's better to call it something like `c` for char since we're iterating on chars or `ubCharIndex` if you feel like it.

tools/src/common/mod.cpp
```diff
@@ -55,6 +57,9 @@ tMod::tMod(const std::string &szFileName)
 
 		// Data read successfully, fill sample info
 		tSample Sample;
+		for (uint8_t i = 0; i < SAMPLE_NAME_SIZE; ++i){ 
+			szSampleNameRaw[i] = toupper(szSampleNameRaw[i]); // sample name to uppercase, to avoid duplicates
```
tehKaiN, line 61: this will work, but the side effect is that the sample names will all be in uppercase in resulting mods. Dunno if it's good - perhaps it's better to leave the case as is and just do name comparison between samples on temporarily uppercased copies.

Also, doesn't it produce error/warning by writing `toupper()` instead of `std::toupper()`? It's better to write it this way.

tehKaiN (changes requested):
please fix the `for` loop iterator name and add the `std::` namespace prefix. I'll leave the decision on uppercase permanence to you.

tehKaiN (approved):

PR #174, opened 2022-12-06
RFC

timfel:


src/ace/managers/viewport/tilebuffer.c
```diff
@@ -219,6 +229,12 @@ void tileBufferReset(
 		}
 	}
 
+	// Init tile offset lookup table
+	pManager->pTileSetOffsets = memAllocFastClear(sizeof(APTR) * 0xFF);
+	for (UBYTE i = 0; i < 0xFF; ++i) {
```
tehKaiN, line 234: ditto - perhaps put the upper limit in some kind of constant and also add ACE_DEBUG-only checks if tile indices later on are smaller?

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -267,11 +283,92 @@ void tileBufferReset(
...
+	} else {
+		// XXX: misuse uwBlitWords to store the flag for non-interleaved
+		uwBlitWords |= UW_BLIT_WORDS_NON_INTERLEAVED_BIT;
+		if (ubSrcInterleaved ^ ubDstInterleaved) {
+			// Since you're using this fn for speed
+			logWrite("WARN: Mixed interleaved - you're losing lots of performance here!\n");
+		}
+		if(ubSrcInterleaved) {
+			wSrcModulo += pManager->pTileSet->BytesPerRow * (pManager->pTileSet->Depth-1);
+		} else if(ubDstInterleaved) {
```
tehKaiN, line 320:
```suggestion
		}
		else if(ubDstInterleaved) {
```

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -267,11 +283,92 @@ void tileBufferReset(
...
+	g_pCustom->bltafwm = 0xFFFF;
+	g_pCustom->bltalwm = 0xFFFF;
+	g_pCustom->bltamod = wSrcModulo;
+	g_pCustom->bltdmod = wDstModulo;
+
+	return (uwHeight << 6) | uwBlitWords;
+}
+
+FN_HOTSPOT
+static inline void tileBufferContinueTileDraw(
```
tehKaiN, line 337: can you measure the `inline` impact? There's a popular belief that compilers tend to ignore it and use its own heuristics for inlining, unless [[always_inline]] is added

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -267,11 +283,92 @@ void tileBufferReset(
...
+	const tTileBufferManager *pManager, UBYTE *pTileDataColumn, UWORD uwTileY,
+	UWORD uwBfrX, UWORD uwBfrY, UWORD uwBltsize, UWORD uwDstBytesPerRow, PLANEPTR pDstPlane
+) {
+	UBYTE ubTileToDraw = pTileDataColumn[uwTileY];
+	ULONG ulDstOffs = uwDstBytesPerRow * uwBfrY + (uwBfrX>>3);
```
tehKaiN, line 342: I think getting rid of this multiplication would speed things up! That kinda seems possible, since for horizontal margins `uwBfrY` is constant and for vertical ones it increments by constant value. Seems like moving that to function above and adding precalculated `uwDstBytesPerRow * ubTileSize` would do the trick!

Or is the compiler smart enough to get rid of it?

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -267,11 +283,92 @@ void tileBufferReset(
...
+
+	if (!(uwBltsize & UW_BLIT_WORDS_NON_INTERLEAVED_BIT)) {
+		UBYTE *pUbBltapt = pManager->pTileSetOffsets[ubTileToDraw];
+		UBYTE *pUbBltdpt = (UBYTE*)((ULONG)pDstPlane + ulDstOffs);
+
+		blitWait(); // Don't modify registers when other blit is in progress
+		g_pCustom->bltapt = pUbBltapt;
+		g_pCustom->bltdpt = pUbBltdpt;
+		g_pCustom->bltsize = uwBltsize;
+	} else {
```
tehKaiN, line 352:
```suggestion
	}
	else {
```

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -267,11 +283,92 @@ void tileBufferReset(
...
+		ULONG ulSrcOffs = (ULONG)pManager->pTileSetOffsets[ubTileToDraw] - (ULONG)pManager->pTileSet->Planes[0];
+		for(UBYTE ubPlane = pManager->pTileSet->Depth; ubPlane--;) {
+			// This hell of a casting must stay here or else large offsets get bugged!
```
tehKaiN, line 355: I kinda wonder if this comment is true - IIRC it was required back when I used VBCC. perhaps GCC is smart enough for adding offset to UBYTE* directly.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -445,6 +567,7 @@ void tileBufferRedrawAll(tTileBufferManager *pManager) {
 		UWORD uwTileOffsX = (wStartX << ubTileShift);
 
 		for (UWORD uwTileX = wStartX; uwTileX < uwEndX; ++uwTileX) {
+			// draw
```
tehKaiN, line 570: indeed

src/ace/managers/viewport/tilebuffer.c
timfel: I looked again, and I just overlooked the `muls.l`. I tried your suggestion, and it shaved another 40 cycles off the time between blits :)

src/ace/managers/viewport/tilebuffer.c
timfel: I removed all those casts, and gcc produces the exact same assembly. I'll leave them out

tehKaiN: care to remove them from managers/blit.c while you're at it? ;)

src/ace/managers/blit.c
```diff
@@ -362,10 +360,10 @@ UBYTE blitUnsafeCopyMask(
 		g_pCustom->bltcmod = wDstModulo;
 		g_pCustom->bltdmod = wDstModulo;
 
-		g_pCustom->bltapt = (UBYTE*)((ULONG)pMsk + ulSrcOffs);
-		g_pCustom->bltbpt = (UBYTE*)((ULONG)pSrc->Planes[0] + ulSrcOffs);
-		g_pCustom->bltcpt = (UBYTE*)((ULONG)pDst->Planes[0] + ulDstOffs);
-		g_pCustom->bltdpt = (UBYTE*)((ULONG)pDst->Planes[0] + ulDstOffs);
+		g_pCustom->bltapt = pMsk + ulSrcOffs;
```
tehKaiN, line 363: whoops! pMsk is UWORD*, so adding hat offset directly won't work. I think it's safe to change that param to UBYTE*.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -299,17 +413,44 @@ void tileBufferProcess(tTileBufferManager *pManager) {
...
+						uwBltsize, ulDstOffs, pDstPlane,
+						// do not set bltdpt, it was left at the right place by the previous blit
+						0
 					);
-					++pState->pMarginX->wTileCurr;
-					uwTileOffsY = (uwTileOffsY + ubTileSize);
-					if(uwTileOffsY >= pManager->uwMarginedHeight) {
-						uwTileOffsY -= pManager->uwMarginedHeight;
+					if (cbTileDraw) {
+						cbTileDraw(uwTilePos, uwTileCurr, pManager->pScroll->pBack, uwTileOffsX, uwTileOffsY);
```
tehKaiN, line 439: it occured to me that this can't be called here if we expect end users to draw stuff on those tiles - calling any other blit function will trash the blitter registers. So this needs to be in a separate loop.

The good news is, it's a speedup for non-callback mode - it can only check if cbTileDraw is non-zero only once, and if so, iterate over new tiles.

tehKaiN:
stuff works reliable so far in Aminer, so I guess we can merge it in its current state?

tehKaiN:
Works stable enough for me. Shall we merge it @timfel ?

tehKaiN (approved):

PR #178, opened 2023-02-08
Copper x wait fix

AdamENESS:
You know, sometimes I forget how to count.
Added an extra element to account for that rare 0 bitplanes..

An alternate solution is to subtract 1 from the bpp when doing the lookup but that would mean it is a little less optimized.


tehKaiN (approved):
Aw crap, and I haven't noticed that! Yeah I think one more byte in data is better than additional instruction bytes here and there.

PR #188, opened 2023-04-28
Misc stuff from ptplayer-update branch

tehKaiN:
Miscellaneous fixes from ptplayer branch which aren't related to ptplayer. Moved to separate branch to make PRs smaller and more digestible.

### DIFF
--- a/include/ace/macros.h
+++ b/include/ace/macros.h
@@ -66,7 +66,8 @@ static inline ULONG ror32(ULONG ulIn, UBYTE ubRot) {
 #define MIN(x,y) ((x)<(y)? (x): (y))
 #define MAX(x,y) ((x)>(y)? (x): (y))
 #define CLAMP(x, min, max) ((x) < (min)? (min) : ((x) > (max) ? (max) : (x)))
-#define ROUND_TO_MULTIPLE(x, m) ((((x) + (m) - 1) / (m)) * (m))
+#define SNAP_UP(x, m) ((((x) + (m) - 1) / (m)) * (m))
+#define SNAP_DOWN(x, m) (((x) / (m)) * (m))
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
 /**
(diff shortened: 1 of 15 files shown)


timfel:
Nothing controversial that I can see. I find the SNAP_UP/DOWN names not great, but don't have a better idea to offer, and they are  still understandable enough to me

tehKaiN:
I saw somewhere snap up/down names recently (don't remember where) and I find it superior to ROUND_TO_MULTIPLE because it's way shorter and I find it has same level of description in its name on what it does.

include/ace/macros.h
```diff
@@ -66,7 +66,8 @@ static inline ULONG ror32(ULONG ulIn, UBYTE ubRot) {
 #define MIN(x,y) ((x)<(y)? (x): (y))
 #define MAX(x,y) ((x)>(y)? (x): (y))
 #define CLAMP(x, min, max) ((x) < (min)? (min) : ((x) > (max) ? (max) : (x)))
-#define ROUND_TO_MULTIPLE(x, m) ((((x) + (m) - 1) / (m)) * (m))
+#define SNAP_UP(x, m) ((((x) + (m) - 1) / (m)) * (m))
+#define SNAP_DOWN(x, m) (((x) / (m)) * (m))
```
approxit, line 70: Any alternatives for macro names?

tehKaiN: Now I remember where I found that name! In Godot 4 there's function called [snapped()](https://docs.godotengine.org/en/latest/classes/class_%40globalscope.html#class-globalscope-method-snapped) which does similar thing, except that it uses typical rounding. In older versions it was called stepify(), which is even less fortunate name. ;)

I think I'd be reasonable to add `#define SNAP(x, m) ((((x) + (m)/2) / (m)) * (m))` for completeness.

tehKaiN: that could work, and be extended to CEIL_TO_FACTOR, FLOOR_TO_FACTOR and ROUND_TO_FACTOR. It's better name than the original one, but it's still a bit long and doesn't group nicely by name, 'cuz the common part is trailing one.

I'll probably use it, since SNAP_* is probably very bad if two separate reviews don't like this name. ;)

@timfel Which ones do you prefer?

timfel: I like the `*_TO_FACTOR` proposal

tehKaiN: It's settled then!

PR #191, opened 2023-05-09
Typo in Managers/Key.h

Vairn:
You seem to have a copy/paste/fail in the key manager, where it thinks it is a mouse.


Vairn:
looks like the request half went through.

At least Github seems to be working again.

PR #194, opened 2023-05-12
Bob manager

### LINKED ISSUE #46: Add basic bob implementation
Currently, bob mechanism must be written from scratch - it should be for most games, but there should be some means to prototype things and I guess that most implementations could have a common ancestor.

It should be mentioned that blitter could be set up using CPU or Copper - the latter option ensures constant blitter throughput.

tehKaiN:
## Description
Adds manager for blitter objects, a common functionality found in almost every game.
Closes #46. 

## Motivation and Context
Managing movable and blittable objects by typical ACE's blitter functions is horribly slow and suboptimal - this manager allows to do it more efficiently.

## ACE inclusion self-assessment

**Are there significantly different solutions to same problem? If so, is it the most common/generic one?**

It can be copper-driven or interrupt-driven queue-based, but this one is the most typical approach.

**Are other solutions better and should ACE incorporate it instead of this change in the long run?**

Other implementations can't be dropped into any codebase. Copper-driven approach could heavily conflict with raw-mode copperlists (e.g. one from scrollbuffer). Interrupt-driven could use code of this implementation, if needed, but it takes performance hit from interrupt-handling overhead.

**Is given functionality needed or nice to have for most of games? Is it used often in tandem with ACE?**

Yes. All my games use it. Almost all of Amiga games use bobs.

**Is it small/simple enough to not require much maintenance?**

Not that small and simple, but quite decoupled - even has its own blitter code.

**Is the functionality related to a core Amiga concept?**

Yes, blittable objects, or bobs.

**Is it entirely, or its ACE interface, written by core ACE team or frequent contributors?**

Yes.

**Is it a building block or dependency for other parts of code? How much coupling there is between this part and rest of ACE code?**

Almost no coupling - only memory allocation etc. I don't see scenario of other ACE things depending on it, unless it's going to be interrupt-driven expansion of the system (perhaps done within the component itself) or some high-level `generic` part of ACE.

**Does it allow making significantly better games at reasonable performance cost?**

Yes. Way better bobs with better performance.

**Is most of the code first-party?**

All of it.

**Does it require additions/changes/removal of some other parts of ACE functionality/tooling for its use?**

No.

## How Has This Been Tested?

Tested across different LMC games:

- Germz
- Chaos Arena
- Aminer (WIP)
- OpenFire (WIP)

## Types of changes

## Checklist

- [ ] My change requires a change to existing documentation.
- [x] My change requires a new chapter in documentation's tutorial.

I'll add docs when I get to the relevant chapter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,9 +49,11 @@ set(ACE_LIBRARY_KINDS "OBJECT" "STATIC")
 set_property(CACHE ACE_LIBRARY_KIND PROPERTY STRINGS ${ACE_LIBRARY_KINDS})
 set(ACE_DEBUG OFF CACHE BOOL "Build with ACE-specific debug/safety functionality.")
 set(ACE_DEBUG_UAE OFF CACHE BOOL "With ACE_DEBUG enabled, output log to UAE console.")
+set(ACE_BOB_WRAP_Y ON CACHE BOOL "Conrols Y-wrapping support in bob manager. Disable for extra performance in simple buffer scenarios.")
 message(STATUS "[ACE] ACE_LIBRARY_KIND: '${ACE_LIBRARY_KIND}'")
 message(STATUS "[ACE] ACE_DEBUG: '${ACE_DEBUG}'")
 message(STATUS "[ACE] ACE_DEBUG_UAE: '${ACE_DEBUG_UAE}'")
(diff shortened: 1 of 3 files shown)


Vairn:
Tested and working with the AGA ACE Branch.
Just make sure you mention that it needs to be used with an INTERLEAVED bitmap for the front and back buffers.

PR #199, opened 2023-07-13
use CMake Package Manager instead of bundling freetype/fmt for tools

tehKaiN:
## Description

Amalgamated version of freetype is problematic due to being outdated (build emitted lots of "register" warnings). Also, some ppl had errors while building tools with OSX, and error messages hinted that this library is a problem.

Instead of bundling freetype and fmt sources, they are now auto-downloaded using [CMake Package Manager (CPM)](https://github.com/cpm-cmake/CPM.cmake). Also, they're now in more recent version and bumping them should be less tedious.

Tested with MS Build Tools 2022 and GCC 12.2 under Windows.

## Checklist


richardleggett:
Just testing this branch on MacOS (x86 and an M1) and both are yielding a new error:

```
In file included from ~/src/ace/tools/src/common/bitmap.cpp:10:
/Users/richardleggett/src/ace/tools/src/common/../common/endian.h:11:17: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
        constexpr bool isBig(void)
                       ^
~/src/ace/tools/src/common/../common/endian.h:18:10: note: read of member 'c' of union with active member 'i' is not allowed in a constant expression
                return uTest.c[0] == 1;
                       ^
1 error generated.
```

If access to those platforms is restricted would be happy to dig into this, maybe with some pointers.

tehKaiN:
huh, looks like clang/gcc mismatch 👀 https://stackoverflow.com/questions/72329759/constexpr-function-returning-member-of-union-g-vs-clang-no-diagnostics-vs

richardleggett:
On a Mac gcc is aliased to clang*, would this be a case we need to force e.g. `gcc-11` via cmake (e.g. using `brew install gcc@11`)?

*
```
gcc --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.3.0
Thread model: posix
InstalledDir: /Applications/Xcode-14.3.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

tehKaiN:
Nah, I prefer fixing it for clang. I've deployed a fix - try now! ;)

richardleggett:
🍾 That got it

PR #207, opened 2023-09-23
Update tools.md instructions for MacOS

richardleggett:
## Description
Out of the box you can't build tools on a Mac. These instructions are a bit transient on purpose since lots of people use a Mac to compile for native Mac/iOS which assumes Clang. If not, you might want to set your system up to always use GCC (or use something like `CMAKE_USER_MAKE_RULES_OVERRIDE` 🤔).

## Checklist
[ ] My code follows the code style of this project.
[x] My change requires a change to the documentation.
[x] I have updated the documentation accordingly.


tehKaiN (approved):
nice!

PR #208, opened 2023-09-23
Extract memLogPeak() into function

richardleggett:
## Description
Extract logging of peak CHIP/RAM usage into callable function in memory.c/h for debugging purposes.

## Checklist

### DIFF
--- a/src/ace/managers/memory.c
+++ b/src/ace/managers/memory.c
@@ -190,10 +190,7 @@ void *_memAllocDbg(
 			"[MEM] ERR: couldn't allocate %lu bytes! (%s:%u)\n",
 			ulSize, szFile, uwLine
 		);
-		logWrite(
-			"[MEM] Peak usage: CHIP: %lu, FAST: %lu\n",
-			s_ulChipPeakUsage, s_ulFastPeakUsage
-		);
+		memLogPeak();
 #ifdef AMIGA
 		logWrite(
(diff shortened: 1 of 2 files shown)


richardleggett:
I'm not 100% sure about the codestyle or ordering so please let me know on those and I'll keep in mind for future.

src/ace/managers/memory.c
```diff
@@ -190,10 +190,7 @@ void *_memAllocDbg(
 			"[MEM] ERR: couldn't allocate %lu bytes! (%s:%u)\n",
 			ulSize, szFile, uwLine
 		);
-		logWrite(
-			"[MEM] Peak usage: CHIP: %lu, FAST: %lu\n",
-			s_ulChipPeakUsage, s_ulFastPeakUsage
-		);
+		memLogPeak();
```
richardleggett, line 193: Should this also stay inlined to avoid a function call?

tehKaiN: it's only for debug builds, so I think a fn call is okay - it won't be fast anyway ;)

tehKaiN (approved):

src/ace/managers/memory.c
richardleggett: That's what I hoped, good to hear. Did I see that CI failed on this, false alarm?

PR #213, opened 2023-11-01
Add support for interleaved blitcopy

tehKaiN:
## Description
Makes the blitCopy() significantly faster when using on interleaved bitmaps.
The showcase twister is faster now.

Also fixes modulo calculations for blitCopyAligned() when using mixed interleaved/non-interleaved bitmaps.

## Checklist


tehKaiN:
What's worrying is that the twister has some odd garbage when switched to interleaved blits. On non-interleaved blits everything is fine. I see no way of reproducing that in isolated test case. 👀

tehKaiN:
works stable enough

PR #216, opened 2023-12-07
add assume.c/.h, convert assert-like error checks into assume statements

### LINKED ISSUE #204: Add ACE-specific assert implementation
- replace `checkNull()` with `assert()` (`aceAssert()`?)
- follow `static_assert()` syntax with 2nd parameter allowing for error message
- should use `logWrite()` and then hang in infinite loop
- in release builds, asserts should do nothing
  - perhaps add a compile option to make them work in release builds with `systemKill()` prints to console?

tehKaiN:
## Description

Adds assertion-like mechanism, which falls back to GCC13's assume attribute. Closes #204. On Bartman suite, this results with a breakpoint on where the failed error check took place.

## Motivation and Context

Getting used to errors or not finding them among logs is quite common. Proposed feature makes them more apparent and also unifies the behavior on failed, often critical, error checks.

It also takes the burden off of failing gracefully of some parts, especially those, which shouldn't ever fail if user code is well formed.

Also generalizes `checkNull()` approach introduced by @approxit in state manager, adding it to remaining part of codebase.

## How Has This Been Tested?

Built Aminer with it, ran the game for a while, so apparently there are no regressions. Did some dummy checks to see if breakpoints on failed assumptions trigger properly.

## Types of changes

## Checklist

### DIFF
--- a/include/ace/utils/assume.h
+++ b/include/ace/utils/assume.h
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef _ACE_UTILS_ASSUME_H_
+#define _ACE_UTILS_ASSUME_H_
+
+#include <ace/types.h>
+#include <ace/managers/system.h>
+#include <ace/managers/log.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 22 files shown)


tehKaiN:
Currently, there's a problem with Bartman suite, which doesn't allow inspecting variables in the middle of the call stack. The dirty solution is ensuring that assertions always trigger one level deep and convert the `_assume()` function into macro.

The better solution would be fixing the debugger support as explained in https://github.com/BartmanAbyss/vscode-amiga-debug/issues/244 . We'll have to see how this pans out.

PR #218, opened 2024-01-11
Enable/disable cdtv device if present on system use/unuse

### LINKED ISSUE #166: CDTV interrupt 2 issue
While working on my game prototype, I might have discovered that when testing on a CDTV (in winUAE), the ACE engine doesn't work:
I've tested using germz ADF from the game's website to make sure it wasn't something specific (since I modified the engine on a few places).
I discovered that int2Handler, which handles the INTF_PORTS interrupts, for CIA A (CIAICRF_SERIAL for keyboard, CIAICRB_TIMER_A for timers), that apparently the CDTV hardware also uses that interrupt for the CD. (When debugging it, it appears as if the int2Handler is triggered continuously, which freezes the ACE engine/game.).

I was able to 'fix' it on my customized ACE engine by disabling INTF_PORTS interrupts and using VERTB to "poll" keyboard. Probably not an ideal solution since perhaps I could miss some keypresses.

I just wanted to at least inform you about the situation. The real proper way to handle it is probably to have an IntServer and let OS handle the CD interrupt instead of masking them.

timfel:
## Description
Closes #166

## Checklist
- [ ] My code follows the code style of this project.

### DIFF
--- a/src/ace/managers/system.c
+++ b/src/ace/managers/system.c
@@ -98,6 +98,8 @@ static const UWORD s_uwOsMinDma = DMAF_DISK | DMAF_BLITTER;
 static struct IOAudio s_sIoAudio = {0};
 static struct Library *s_pCiaResource[CIA_COUNT];
 static struct Process *s_pProcess;
+static struct MsgPort *s_pCdtvIOPort;
+static struct IOStdReq *s_pCdtvIOReq;
 
 #if defined(BARTMAN_GCC)
 struct DosLibrary *DOSBase = 0;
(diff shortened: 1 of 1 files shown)


src/ace/managers/system.c
```diff
@@ -342,6 +344,25 @@ static void msgPortDelete(struct MsgPort *mp) {
...
+static struct IORequest *ioRequestCreate(struct MsgPort *pMsgPort, UWORD uwLen) {
+	struct IORequest *pIoReq = (struct IORequest *) AllocMem(uwLen, MEMF_PUBLIC | MEMF_CLEAR);
+	if (!pIoReq) {
+		return NULL;
+	}
+	ioRequestInitialize(pIoReq, pMsgPort, uwLen);
+	return pIoReq;
+}
+
+static void ioRequestDelete(struct IORequest *pIoReq) {
```
tehKaiN, line 362: I think this will be more consistent with ACE naming scheme
```suggestion
static void ioRequestDestroy(struct IORequest *pIoReq) {
```

src/ace/managers/system.c
```diff
@@ -593,6 +649,15 @@ void systemUnuse(void) {
 
 	--s_wSystemUses;
 	if(!s_wSystemUses) {
+		// Disable CDTV CD drive if present
```
tehKaiN, line 652: Perhaps this should be after systemFlushIo() 'cuz I think it works for all floppy/hdd/cd I/O, but I'm not quite sure.

tehKaiN (approved):
Looks good!

src/ace/managers/system.c
timfel: done

timfel:
Not sure why the CI doesn't work, since it's not public :)

PR #220, opened 2024-04-19
generic main: add ability to disable timer manager

tehKaiN:
## Description

Adds ability to disable timer manager in generic main code.

I've left it on by default, because currently log blocks depend on it to display time measurements (like they were correct anyway lol). Also, this approach won't break already made games.

## Checklist


tehKaiN:
no comments? Ok, let's merge it then.

PR #247, opened 2025-02-03
Change sprite manager to allow user to pass their own blank sprite.

timfel:
This is not nice, but I wanted to start discussing how to get around the problem that I get into conflicts with the sprite manager if I do my own sprite system:

```
[build] D:/greatflight/src/sprites.c:126:19: error: 's_spriteData' causes a section type conflict with 's_uBlankSprite'
[build]   126 | static UWORD CHIP s_spriteData[] = {
[build]       |                   ^
[build] D:/greatflight/deps/ACE/src/ace/managers/sprite.c:29:41: note: 's_uBlankSprite' was declared here
[build]    29 | static const tHardwareSpriteHeader CHIP s_uBlankSprite;
[build]       |                                         ^
[build] lto-wrapper.exe: fatal error: c:\Users\timfe\.vscode\extensions\bartmanabyss.amiga-debug-1.7.7\bin\win32\opt\bin\m68k-amiga-elf-gcc.exe returned 1 exit status
```

This hack works around it for me and for now, but it's confusing since it doesn't actually prevent you from using the sprite system and then you get error messages that are again hard to understand. So just putting this here to discuss.

### DIFF
--- a/docs/programming/sprites.md
+++ b/docs/programming/sprites.md
@@ -49,8 +49,8 @@ The manager works in block, as well as raw copperlist mode.
 void myGsCreate(void) {
   // Some stuff goes here...
 
-  spriteManagerCreate(s_pView, 0); // For raw copper mode, pass position on
-                                   // copperlist for sprite initialization.
+  spriteManagerCreate(s_pView, 0, NULL); // For raw copper mode, pass position on
+                                         // copperlist for sprite initialization.
   systemSetDmaBit(DMAB_SPRITE, 1); // Enable sprite DMA.
 
   // Stuff continues...
--- a/include/ace/managers/sprite.h
+++ b/include/ace/managers/sprite.h
@@ -48,13 +48,15 @@ typedef struct tSprite {
  * @param pView View used for displaying sprites.
  * @param uwRawCopPos In raw mode, specifies an offset on where
  * the sprite commands should reside. Requires space of 16 copper commands.
+ * @param pBlankSprite 2 words of CHIP memory (the blank sprite control words).
+ * Pass NULL to let the manager deal with the blank sprite memory.
  *
  * @see spriteDisableInCopBlockMode()
  * @see spriteDisableInCopRawMode()
  * @see systemSetDmaBit()
  * @see spriteManagerDestroy()
  */
-void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos);
+void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos, ULONG pBlankSprite[1]);
 
 /**
  * @brief Destroys the hardware sprite manager.
--- a/include/ace/utils/sprite.h
+++ b/include/ace/utils/sprite.h
@@ -26,10 +26,11 @@ extern "C" {
  * @param pList Copperlist to be edited.
  * @param eSpriteMask Determines sprites to be disabled.
  * @param uwCmdOffs Start position on raw copperlist.
+ * @param pBlankSprite 2 words of CHIP memory (the blank sprite control words)
  * @return Number of MOVE instructions added.
  */
 UBYTE spriteDisableInCopRawMode(
-	tCopList *pList, tSpriteMask eSpriteMask, UWORD uwCmdOffs
+	tCopList *pList, tSpriteMask eSpriteMask, UWORD uwCmdOffs, ULONG pBlankSprite[1]
 );
 
 /**
@@ -38,9 +39,10 @@ UBYTE spriteDisableInCopRawMode(
  *
  *  @param pList Copperlist to be edited.
  *  @param eSpriteMask Determines sprites to be disabled.
+ *  @param pBlankSprite 2 words of CHIP memory (the blank sprite control words)
  *
  *  @return Pointer to newly created copBlock.
  */
-tCopBlock *spriteDisableInCopBlockMode(tCopList *pList, tSpriteMask eSpriteMask);
+tCopBlock *spriteDisableInCopBlockMode(tCopList *pList, tSpriteMask eSpriteMask, ULONG pBlankSprite[1]);
 
 #endif // _ACE_UTILS_SPRITE_H_
--- a/src/ace/managers/sprite.c
+++ b/src/ace/managers/sprite.c
@@ -44,7 +44,7 @@ void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos, ULONG pBlankSpri
 		s_isOwningBlankSprite = 1;
 		s_pBlankSprite = memAllocChipClear(sizeof(ULONG));
 		// Just to make sure we don't accidentally mismatch the control words size
-		_Static_assert(sizeof(ULONG) == sizeof(tHardwareSpriteHeader));
+		_Static_assert(sizeof(ULONG) == sizeof(tHardwareSpriteHeader), "We expect a Hardware sprite to have a ULONG sized header");
 	}
 	// TODO: add support for non-chained mode (setting sprxdat with copper)?
 	s_pView = pView;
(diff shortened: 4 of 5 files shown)


tehKaiN:
Damn, that's one puzzling error. Not sure why it happens - it's probably because `CHIP` thingy not being properly handled by Bartman's compiler and going nuts when more than one symbol (or compilation unit) uses it. Perhaps it's for the best to not use it at all and refactor the API so that this empty buffer would be supplied by user.

src/ace/managers/sprite.c
```diff
@@ -23,14 +23,29 @@ typedef struct tSpriteChannel {
...
 static tCopBlock *s_pInitialClearCopBlock;
+ULONG *s_pBlankSprite;
 
 static void spriteChannelRequestCopperUpdate(tSpriteChannel *pChannel) {
 	pChannel->ubCopperRegenCount = 2; // for front/back buffers in raw mode
 }
 
-void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos) {
+void spriteManagerCreate1(const tView *pView, UWORD uwRawCopPos) {
+	s_pBlankSprite = memAllocChipClear(sizeof(ULONG));
```
timfel, line 35: ofc if we dynamically allocate it anyway, we might as well just not allow the user to manage it and always allocate dynamically. the problem that bartman's gcc has is simply that during lto it seems to want to merge all CHIP static variables and then I cannot have a custom CHIP static variable in my program that is of a different storage type than `tHardwareSpriteHeader`.

include/ace/utils/sprite.h
```diff
@@ -17,6 +17,8 @@ extern "C" {
 
 #include <ace/managers/copper.h>
 
+extern ULONG *s_pBlankSprite;
```
tehKaiN, line 20: if that's an extern, it's no longer `s_`, must be a `g_`

include/ace/managers/sprite.h
```diff
@@ -45,16 +49,20 @@ typedef struct tSprite {
  * @note This function doesn't handle the mouse input etc. automatically,
  * since one may want to control it with joy, keyboard or in other kind of way.
  *
- * @param pView View used for displaying sprites.
- * @param uwRawCopPos In raw mode, specifies an offset on where
+ * @param const tView *pView View used for displaying sprites.
```
tehKaiN, line 52: hmmm, all other doxycomments in ACE skip parameter types, and I'm not sure which way is better

src/ace/managers/sprite.c
```diff
@@ -23,14 +23,29 @@ typedef struct tSpriteChannel {
...
+	spriteManagerCreate2(pView, uwRawCopPos, s_pBlankSprite);
+	s_ubOwnsBlankSprite = 1;
+}
+
+void spriteManagerCreate2(const tView *pView, UWORD uwRawCopPos, ULONG *pBlankSprite) {
```
tehKaiN, line 40: how about only having this function and if pBlankSprite is null then allocate and set s_ubOwnsBlankSprite to 1?

src/ace/managers/sprite.c
```diff
@@ -23,14 +23,29 @@ typedef struct tSpriteChannel {
...
+#ifdef ACE_DEBUG
+	if ((ULONG)pBlankSprite > memGetChipSize() - sizeof(ULONG)) {
+		logWrite("ERR: ILLEGAL NON-CHIP memory location for blank sprite!");
+	}
+#endif
+	s_ubOwnsBlankSprite = 0;
```
tehKaiN, line 46: `s_isOwningBlankSprite`?

src/ace/utils/sprite.c
```diff
@@ -21,7 +19,7 @@ tCopBlock *spriteDisableInCopBlockMode(tCopList *pList, tSpriteMask eSpriteMask)
 	}
 
 	// Set instructions
-	ULONG ulBlank = s_uBlankSprite.ulRaw;
+	ULONG ulBlank = (ULONG)s_pBlankSprite;
```
tehKaiN, line 22: my biggest problem with this change is that if someone just wants to use utils/sprite.h and not sprite manager, then there's no indicator that s_pBlankSprite needs to be set to valid alloc prior to spriteDisableInCopBlockMode. I see 2 solutions:

1. pass the blank sprite buffer through param and remove the global altogether
2. specify in doxycomment that it needs to be initialized first (but many ppl skip docs) and add null check (but var might be set to random value depending on how it's declared).

I'd personally lean towards 1 since it gives user less choice to break things.

timfel:
@tehKaiN I went with your suggestion in https://github.com/AmigaPorts/ACE/pull/247#discussion_r1942392753 to always pass the blank sprite buffer to the utils/sprite functions.

docs/programming/sprites.md
```diff
@@ -49,8 +49,8 @@ The manager works in block, as well as raw copperlist mode.
 void myGsCreate(void) {
   // Some stuff goes here...
 
-  spriteManagerCreate(s_pView, 0); // For raw copper mode, pass position on
-                                   // copperlist for sprite initialization.
+  spriteManagerCreate(s_pView, 0, NULL); // For raw copper mode, pass position on
```
tehKaiN, line 52: we don't use NULL too much in ACE (usually zero) but now I think we should do that more often

src/ace/managers/sprite.c
```diff
@@ -23,14 +23,29 @@ typedef struct tSpriteChannel {
...
+			logWrite("ERR: ILLEGAL NON-CHIP memory location for blank sprite!");
+		}
+#endif
+		s_isOwningBlankSprite = 0;
+		s_pBlankSprite = pBlankSprite;
+	} else {
+		s_isOwningBlankSprite = 1;
+		s_pBlankSprite = memAllocChipClear(sizeof(ULONG));
+		// Just to make sure we don't accidentally mismatch the control words size
+		_Static_assert(sizeof(ULONG) == sizeof(tHardwareSpriteHeader));
```
tehKaiN, line 47: can you by any chance check if static asserts are supported by Bebbo compiler? I don't have one at hand currently.

tehKaiN (approved):

include/ace/managers/sprite.h
```diff
@@ -48,13 +48,15 @@ typedef struct tSprite {
...
+ * @param pBlankSprite 2 words of CHIP memory (the blank sprite control words).
+ * Pass NULL to let the manager deal with the blank sprite memory.
  *
  * @see spriteDisableInCopBlockMode()
  * @see spriteDisableInCopRawMode()
  * @see systemSetDmaBit()
  * @see spriteManagerDestroy()
  */
-void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos);
+void spriteManagerCreate(const tView *pView, UWORD uwRawCopPos, ULONG pBlankSprite[1]);
```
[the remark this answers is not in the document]
timfel: I was going back and forth on this. I like that `[1]` gives a hint to the caller and we can pass a `ULONG*` without warnings anyway, because of the whole pointer decay rules stuff. I always find myself wondering with these pointer APIs where I know "they want an array, but how many elements?". (I like `ULONG pBlankSprite[static 1]` even more because it gives a warning when I pass not enough elements, but then passing `NULL` warns, too. It's more useful when expecting arrays with a specific element count anyway).

I suppose a pointer is ok, because a caller can assume the expected size is at least `1` element. So, change it or not?

docs/programming/sprites.md
timfel: Yeah, I like `NULL` way more just because it communicates to me I'm passing this for a pointer, not an integral

tehKaiN (changes requested):
the memory type check is a blocker which I missed early on

src/ace/managers/sprite.c
timfel: I kicked off a Bebbo install and will see if it works

timfel: Ok, Bebbo's gcc version did not like `_Static_assert` without a message, but now it's ok.

tehKaiN (approved):

PR #248, opened 2025-02-04
Update wait positions for EHB and AGA from experience.

timfel:
This encodes the experience of @timfel and @Vairn until we add proper calculation.


tehKaiN (approved):
This is getting one hell of the comment for a single const ;)

PR #257, opened 2025-06-24
Add `-nocs` option to skip OCS color checks

petmac:
## Description
Adds a new optional command line argument, `-nocs`, which skips the check that each color in the palette can be truncated to OCS limits (12-bit color) without loss of data.

## Checklist

- [x] My change requires a change to the documentation.


tehKaiN (approved):
Looks good! I'd probably also update the cmake function for invoking the tool, but that can be added when someone needs it.

PR #258, opened 2025-06-25
Add CONVERT_COLORS to convertPalette CMake function

petmac:
## Description
Add CONVERT_COLORS to convertPalette CMake function. Also update docs.

I don't have much CMake experience so it was mostly copied and pasted from how convertBitmaps does it.

## Checklist

- [x] My change requires a change to the documentation.

### DIFF
--- a/cmake/ace_functions.cmake
+++ b/cmake/ace_functions.cmake
@@ -32,9 +32,20 @@ endfunction()
 
 function(convertPalette TARGET PALETTE_IN PALETTE_OUT)
 	getToolPath(palette_conv TOOL_PALETTE_CONV)
+	set(options CONVERT_COLORS)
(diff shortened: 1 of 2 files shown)


tehKaiN (changes requested):
Damn, I messed up with previous merge. I tend to default to positive logic (e.g. name a switch DO_SOMETHING instead of DONT_DO_SOMETHING) and this one's starting a precedent in the other way.

Could you please rename `NO_FORCE_OCS` to `CONVERT_COLORS` and the switch from `-nocs` to `-cc` (Convert colors to OCS limitations instead of failing on non-compliant ones)? My first idea was `CONVERT_TO_OCS` but that would mean we'd have `-cocs` lol.

PR #262, opened 2025-09-30
Tilebuffer redraw optimizations

timfel:
This is as fast I seem to be able to get it for interleaved full redraws.

image
image

### DIFF
--- a/include/ace/managers/system.h
+++ b/include/ace/managers/system.h
@@ -94,6 +94,23 @@ UWORD systemGetVersion(void);
 
 UBYTE systemIsStartVolumeWritable(void);
 
+/**
+ * Disable caches on 680x0 CPUs. Previous cache control words
+ * are stored so they can later be restored with
+ * systemRestoreCpuCaches.
+ *
+ * @see systemRestoreCpuCaches
+ */
+void systemDisableCpuCaches();
+
+/**
+ * Restore cache control settings after a previous call to
+ * systemDisableCpuCaches.
+ *
+ * @see systemDisableCpuCaches
+ */
+void systemRestoreCpuCaches();
+
 //---------------------------------------------------------------------- GLOBALS
 
 extern struct GfxBase *GfxBase;
--- a/include/ace/managers/viewport/tilebuffer.h
+++ b/include/ace/managers/viewport/tilebuffer.h
@@ -195,19 +195,18 @@ void tileBufferReset(
 
 /**
  * Redraws tiles on whole screen.
- * Use for init or something like that, as it's slooooooooow.
- * Be sure to have display turned off or palette dimmed since even on double
- * buffering it will redraw both buffers.
+ * 
+ * This method is slower than redrawing only what's needed, but depending on
+ * how many tiles need to be redrawn every frame, it may be competitive.
+ * If you only use it for initialization, have display turned off or palette dimmed
+ * since it will redraw both buffers for optimal performance.
+ * 
+ * When a lot is going on (many BOBs), the code when used with interleaved
+ * tilebuffer and tilemap is pretty optimized, to the point that redraw of half a
+ * 4bpp low-res screen of 16x16 tiles is possible on an A500. That saves then having
+ * to undraw BOBs, and may thus be better und some circumstances.
  */
-void tileBufferRedrawAllNonInterleaved(tTileBufferManager *pManager);
-
-/**
- * Redraws tiles on whole screen.
- * Use for init or something like that, as it's slooooooooow.
- * Be sure to have display turned off or palette dimmed since even on double
- * buffering it will redraw both buffers.
- */
-void tileBufferRedrawAllInterleaved(tTileBufferManager *pManager);
+void tileBufferRedrawAll(tTileBufferManager *pManager);
 
 /**
  * Redraws selected tile, calls custom redraw callback
--- a/include/ace/managers/viewport/tilebuffer.h
+++ b/include/ace/managers/viewport/tilebuffer.h
@@ -199,7 +199,15 @@ void tileBufferReset(
  * Be sure to have display turned off or palette dimmed since even on double
  * buffering it will redraw both buffers.
  */
-void tileBufferRedrawAll(tTileBufferManager *pManager);
+void tileBufferRedrawAllNonInterleaved(tTileBufferManager *pManager);
+
+/**
+ * Redraws tiles on whole screen.
+ * Use for init or something like that, as it's slooooooooow.
+ * Be sure to have display turned off or palette dimmed since even on double
+ * buffering it will redraw both buffers.
+ */
+void tileBufferRedrawAllInterleaved(tTileBufferManager *pManager);
 
 /**
  * Redraws selected tile, calls custom redraw callback
--- a/include/ace/managers/viewport/tilebuffer.h
+++ b/include/ace/managers/viewport/tilebuffer.h
@@ -196,17 +196,27 @@ void tileBufferReset(
 /**
  * Redraws tiles on whole screen.
  * 
+ * Use for init or something like that, as it's slooooooooow.
+ * Be sure to have display turned off or palette dimmed since even on double
+ * buffering it will redraw both buffers.
+ * 
+ * If you want to only fully redraw the back buffer, use tileBufferRedrawBack.
+ *
+ * @see tileBufferRedrawBack
+ */
+void tileBufferRedrawAll(tTileBufferManager *pManager);
+
+/**
+ * Redraws the tiles on the entire backbuffer.
+ * 
  * This method is slower than redrawing only what's needed, but depending on
  * how many tiles need to be redrawn every frame, it may be competitive.
- * If you only use it for initialization, have display turned off or palette dimmed
- * since it will redraw both buffers for optimal performance.
- * 
  * When a lot is going on (many BOBs), the code when used with interleaved
  * tilebuffer and tilemap is pretty optimized, to the point that redraw of half a
  * 4bpp low-res screen of 16x16 tiles is possible on an A500. That saves then having
  * to undraw BOBs, and may thus be better und some circumstances.
  */
-void tileBufferRedrawAll(tTileBufferManager *pManager);
+void tileBufferRedrawBack(tTileBufferManager *pManager);
 
 /**
  * Redraws selected tile, calls custom redraw callback
--- a/src/ace/managers/system.c
+++ b/src/ace/managers/system.c
@@ -1189,7 +1189,7 @@ void systemReleaseBlitterToOs(void) {
 }
 
 UBYTE systemBlitterIsUsed(void) {
-	return s_wSystemBlitterUses > 0;
+	return s_wSystemBlitterUses <= 0;
 }
 
 void systemSetKeyInputHandler(tKeyInputHandler cbKeyInputHandler) {
--- a/src/ace/managers/system.c
+++ b/src/ace/managers/system.c
@@ -59,8 +59,33 @@ typedef struct _tAceInterrupt {
 
 // Store VBR query code in .text so that it plays nice with instruction cache
 // TODO: move to .s file after vasm support gets merged into cmake
+// Instructions: movec vbr,d0; rte
 __attribute__((section("text")))
 static const UWORD s_pGetVbrCode[] = {0x4e7a, 0x0801, 0x4e73};
+
+// Get cache control register contents (020+).
+// Instructions: movec cacr,d0; rte
+__attribute__((section("text")))
+static const UWORD s_pGetCacr[] = {0x4e7a, 0x0002, 0x4e73};
+
+// Get processor configuration register contents (060+).
+// Instructions: movec pcr,d0; rte
+__attribute__((section("text")))
+static const UWORD s_pGetPcr[] = {0x4e7a, 0x0801, 0x4e73};
+
+// Move d0 into the cache control register
+// movec d0,cacr; rte
+__attribute__((section("text")))
+static const UWORD s_pSetCacr[] = {0x4e7b, 0x0002, 0x4e73};
+
+// Move d0 into the cache control register and flush caches
+// movec d0,cacr; cpusha ic+dc; rte
+__attribute__((section("text")))
+static const UWORD s_pSetCacr040[] = {0x4e7b, 0x0002, 0xf4f8, 0x4e73};
+
+// Move d0 into the processor configuration register
+__attribute__((section("text")))
+static const UWORD s_pSetPcr[] = {0x4e7b, 0x801, 0x4e73};
 
 // Saved regs
 static UWORD s_uwOsIntEna;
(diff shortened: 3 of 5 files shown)


src/ace/managers/viewport/tilebuffer.c
```diff
@@ -389,8 +393,16 @@ static inline void tileBufferContinueTileDraw(
 			// folds away
 			pUbBltdpt = pDstPlane + ulDstOffs;
 		}
-
-		blitWait(); // Don't modify registers when other blit is in progress
+#ifndef ACE_DEBUG
+		if (ubBlitWait) {
+#else
+		if (1) {
+#endif
```
tehKaiN, line 400: I think I'd might do

```
#if !defined(ACE_DEBUG)
#define TILEBUFFER_REDRAW_HOG 1
#else
#define TILEBUFFER_REDRAW_HOG 0
#endif
```

at the top of the file, do the `if(!TILEBUFFER_REDRAW_HOG || isWaitForBlit)` so that the actual code is less cluttered by preprocessor. Some other checks for `#if defined(ACE_DEBUG)` can be replaced with `if(!TILEBUFFER_REDRAW_HOG)` then.

It's a matter of style really, I'm open to discussion.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -626,7 +648,9 @@ void tileBufferProcess(tTileBufferManager *pManager) {
 	pManager->ubStateIdx = !pManager->ubStateIdx;
 }
 
-void tileBufferRedrawAll(tTileBufferManager *pManager) {
+// This is used below with ubNonInterleavedBlit either 0 or 1, so with the inlining this is
+// a poor man's function template specialization...
+static inline void tileBufferRedrawAllInternal(tTileBufferManager *pManager, UBYTE ubNonInterleavedBlit) {
```
tehKaiN, line 653: I'm 95% sure that `inline` isn't reliable by itself, we should go with `__attribute__((always_inline))`. Bartman's debugger works fine with it btw.

Rename `ubNonInterleavedBlit` to `isInterleaved` and reverse its logic, not a fan of `isNotDoingSomething` or `isNotBeingSomething`, unless it's performance critical, e.g. skips cmp because of it being defaulted to zero in the fast path.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -658,24 +682,60 @@ void tileBufferRedrawAll(tTileBufferManager *pManager) {
...
-	UWORD uwDstOffsStep = ubTileSize / 8;
+	UWORD uwMarginedHeight = pManager->uwMarginedHeight;
+	UWORD uwWrapAroundY = MIN(uwEndY, uwMarginedHeight >> ubTileShift);
+	UWORD uwBfrY = wStartY << ubTileShift;
 	systemSetDmaBit(DMAB_BLITHOG, 1);
-	for (UWORD uwTileY = wStartY; uwTileY < uwEndY; ++uwTileY) {
-		UWORD uwTileCurr = wStartX;
-		ULONG ulDstOffs = uwDstBytesPerRow * uwTileOffsY + uwTileOffsX / 8;
-		while(uwTileCurr < uwEndX) {
+	Disable(); // Disable and clear CPU caches (if any)
```
tehKaiN, line 698: I'm not sure if `Disable()`/`Enable()` is proper way to handle it, especially that it also touches multitasking stuff according to docs and we might have OS dead on our end.

This is tricky, because I'd see it as `systemDisableCache()` that would do `Disable()` when OS is on, and some inline assembly when OS is off. Or, we could assume that ownblit is good enouhh with OS on and just use the inline asm to disable caches. We could also make a debug check for OS being turned off as a prerequisite for hot path.

In the past, I had some stability issues when using Forbid/Permit or Disable/Enable but that was a long ago and I don't remember details. Docs say they shouldn't be used for too long, and I'm not sure how long is too long.

include/ace/managers/viewport/tilebuffer.h
```diff
@@ -199,7 +199,15 @@ void tileBufferReset(
  * Be sure to have display turned off or palette dimmed since even on double
  * buffering it will redraw both buffers.
  */
-void tileBufferRedrawAll(tTileBufferManager *pManager);
+void tileBufferRedrawAllNonInterleaved(tTileBufferManager *pManager);
```
tehKaiN, line 202: If I read this correctly, then all functions are fine to call at any time because they don't require extra flags to work, besides interleaved.

The doc comment needs some info what are the tradeoffs between them so that user can make an informed decision.

The typical ACE way is to either pass a tag to create that switches the behavior, or setting a define that rewires the code. I'm not sure how I feel about that third way, especially that user will probably only go with one of those two ways exclusively on a per-project basis. I need some more time/opinions on this to tell if it's ok or not.

tehKaiN (changes requested):
Looks great, but needs some extra work

include/ace/managers/viewport/tilebuffer.h
tehKaiN: Sure! I guess we can also take the performance hit from the `if (ubSrcInterleaved && ubDstInterleaved) {` as it's probably negligible in total frame time. Or if it proves problematic sometime we can cache it in a variable or something.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -376,11 +376,15 @@ static UWORD tileBufferSetupTileDraw(const tTileBufferManager *pManager) {
...
 	const tTileBufferManager *pManager, const tTileBufferTileIndex *pTileDataColumn,
-	UWORD uwTileY, UWORD uwBltsize, ULONG ulDstOffs, PLANEPTR pDstPlane, UBYTE ubSetDst
+	UWORD uwTileY, UWORD uwBltsize, ULONG ulDstOffs, PLANEPTR pDstPlane, UBYTE ubSetDst,
+	UBYTE ubBlitWait, UBYTE ubNonInterleavedBlit
 ) {
 	tTileBufferTileIndex TileToDraw = pTileDataColumn[uwTileY];
 
-	if (!(uwBltsize & BLIT_WORDS_NON_INTERLEAVED_BIT)) {
+	if (!ubNonInterleavedBlit) {
+		// this function should be inlined into the caller, where
```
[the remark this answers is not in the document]
timfel: Done

src/ace/managers/viewport/tilebuffer.c
timfel: Done

src/ace/managers/viewport/tilebuffer.c
timfel: Done

include/ace/managers/viewport/tilebuffer.h
timfel: Pushed with the small amount of duplicated code.

src/ace/managers/viewport/tilebuffer.c
timfel: This one I still need to think about. Also, `CacheClearU` seems to be a KS2 function, so that's also a problem.

tehKaiN (commented):
Mostly looks good, some nitpicks, but some questions need to be answered before definitive approval. 

That assembly isn't quite readable due to dc.l and stuff like `jsr -0x7e(%%a6)` but there's very little we can do about it. We could put that in separate .s file, but that would either mean using same asm syntax flavor as Bartman's .s files or making people set up vasm properly for cmake to use. It can also stay the way as is for now.

src/ace/managers/system.c
```diff
@@ -1299,3 +1308,125 @@ UBYTE systemIsStartVolumeWritable(void) {
...
+void systemDisableCpuCaches() {
+	++s_cpuCacheFlags.wUses;
+	if (s_cpuCacheFlags.wUses > 1) {
+		return;
+	}
+
+	// Adapted from https://github.com/keirf/flashfloppy-autoswap/blob/master/shared/disable_caches.S
+	// which itself is based on original code from ross at https://eab.abime.net/showpost.php?p=1303326&postcount=3
+    __asm__ volatile(
+        "movem.l %%a3-%%a6/%%d0-%%d1,-(%%sp)\n\t"   /* Save caller registers */
```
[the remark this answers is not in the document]
timfel: Done

src/ace/managers/system.c
```diff
@@ -1299,3 +1308,125 @@ UBYTE systemIsStartVolumeWritable(void) {
...
+        "movea.l %0,%%a3\n\t"                       /* A3 = tCPuCacheFlags* argument */
+        "lea     0x10.w,%%a4\n\t"                   /* A4 = Illegal Instruction vector */
+        "lea     1f(%%pc),%%a5\n\t"   		        /* A5 = Illegal instruction trap handler */
+        "movea.l 4.w,%%a6\n\t"                      /* A6 = ExecBase */
+        "moveq   #0,%%d0\n\t"				        /* D0 = 0 */
+
+        "jsr     -0x78(%%a6)\n\t"                   /* Exec Disable() */
```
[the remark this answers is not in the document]
timfel: I pulled those calls out of the assembly and check `systemIsUsed` around them

src/ace/managers/system.c
```diff
@@ -100,6 +100,14 @@ static tKeyInputHandler s_cbKeyInputHandler;
 static WORD s_wSystemUses;
 static WORD s_wSystemBlitterUses;
 
+struct tCpuCacheFlags {
```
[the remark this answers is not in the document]
timfel: Done

src/ace/managers/system.c
```diff
@@ -955,6 +955,16 @@ void systemDestroy(void) {
 	g_pCustom->dmacon = 0x07FF;
 	g_pCustom->intreq = 0x7FFF;
 
+	if(s_sCpuCacheFlags.wUses != 0) {
+		if(s_sCpuCacheFlags.ulCacr == 0 && s_sCpuCacheFlags.ulPcr == 0) {
+			if(s_sCpuCacheFlags.wUses != 1) {
```
tehKaiN, line 960: That's good way of solving it!

src/ace/managers/system.c
```diff
@@ -1310,123 +1320,138 @@ UBYTE systemIsStartVolumeWritable(void) {
...
 void systemDisableCpuCaches() {
-	++s_cpuCacheFlags.wUses;
-	if (s_cpuCacheFlags.wUses > 1) {
+	++s_sCpuCacheFlags.wUses;
+	if (s_sCpuCacheFlags.wUses > 1) {
 		return;
 	}
 
+	if (systemIsUsed()) {
+		Disable();
```
tehKaiN, line 1329: Love this solution!

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -693,7 +693,7 @@ static inline void tileBufferRedrawAllInternal(tTileBufferManager *pManager, UBY
 	UWORD uwMarginedHeight = pManager->uwMarginedHeight;
 	UWORD uwWrapAroundY = MIN(uwEndY, uwMarginedHeight >> ubTileShift);
 	systemSetDmaBit(DMAB_BLITHOG, 1);
-	Disable(); // Disable and clear CPU caches (if any)
+	systemGetBlitterFromOs();
```
tehKaiN, line 696: This isn't really needed, ACE defaults to blitter ownership and only gives it back to OS for file IO. Should've documented those functions and explain this.

If you're paranoid, you can add a check if blitter is owned by ACE using `systemBlitterIsUsed()` and print an error if it's not.

src/ace/managers/viewport/tilebuffer.c
```diff
@@ -658,24 +681,59 @@ void tileBufferRedrawAll(tTileBufferManager *pManager) {
...
 	UWORD uwTileOffsX = (wStartX << ubTileShift);
-	UWORD uwDstOffsStep = ubTileSize / 8;
+	UWORD uwMarginedHeight = pManager->uwMarginedHeight;
+	UWORD uwWrapAroundY = MIN(uwEndY, uwMarginedHeight >> ubTileShift);
 	systemSetDmaBit(DMAB_BLITHOG, 1);
-	for (UWORD uwTileY = wStartY; uwTileY < uwEndY; ++uwTileY) {
-		UWORD uwTileCurr = wStartX;
-		ULONG ulDstOffs = uwDstBytesPerRow * uwTileOffsY + uwTileOffsX / 8;
-		while(uwTileCurr < uwEndX) {
+	systemGetBlitterFromOs();
```
tehKaiN, line 696: This isn't needed since ACE defaults to having blitter ownership - those functions are only used for file IO to make sure that OS has access to it, because it needs it in some cases.

I should've added comments explaining that for those functions.

If you're paranoid, you can add a check using `systemBlitterIsUsed()` if ACE has blitter ownership at this point and if not, print an `ERR:`.

src/ace/managers/system.c
```diff
@@ -946,6 +955,16 @@ void systemDestroy(void) {
 	g_pCustom->dmacon = 0x07FF;
 	g_pCustom->intreq = 0x7FFF;
 
+	if(s_sCpuCacheFlags.wUses != 0) {
+		if(s_sCpuCacheFlags.ulCacr == 0 && s_sCpuCacheFlags.ulPcr == 0) {
+			if(s_sCpuCacheFlags.wUses != 1) {
+				logWrite("ERR: unclosed/overclosed cpu cache count: %hd\n", s_sCpuCacheFlags.wUses - 1);
```
tehKaiN, line 961: That's a great way to solve the usage problem!

src/ace/managers/system.c
```diff
@@ -1299,3 +1318,140 @@ UBYTE systemIsStartVolumeWritable(void) {
...
 }
+
+void systemDisableCpuCaches() {
+	++s_sCpuCacheFlags.wUses;
+	if (s_sCpuCacheFlags.wUses > 1) {
+		return;
+	}
+
+	if (systemIsUsed()) {
+		Disable();
```
tehKaiN, line 1329: Love this!

tehKaiN (approved):
Looks good. Have you given this a spin in your game to see how stable/reliably it works?

include/ace/managers/system.h
```diff
@@ -94,6 +94,23 @@ UWORD systemGetVersion(void);
...
+ */
+void systemDisableCpuCaches();
+
+/**
+ * Restore cache control settings after a previous call to
+ * systemDisableCpuCaches.
+ *
+ * @see systemDisableCpuCaches
+ */
+void systemRestoreCpuCaches();
```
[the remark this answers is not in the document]
tehKaiN: Fair point

tehKaiN:
Could be like that if you want to pursue it. I think you can query attnflags like I do with VBR to get the CPU model, not sure if it goes up to 060 though.

src/ace/managers/viewport/tilebuffer.c
timfel: Changed the contents :)

tehKaiN (approved):
Works fine with Aminer and even still fits the floppy!

PR #266, opened 2026-04-12
fix display register set order for tSimpleBuffer

tehKaiN:
## Description

This makes the initialization order in-line with `tScrollBuffer`. Fixes graphics glitches on non-emulated AGA, notably in GermZ Survivor.

Also adds the proper copperlist termination with room for double final WAIT cmds.

## Checklist


timfel:
Tested just now, and this makes my bobs flicker like crazy. I did verify by just switching between this one commit and the one before, this one does that. Not sure why, I will try to find time to investigate tomorrow

tehKaiN:
This sounds like double buffering doesn't work as intended, making the game display the backbuffer. Not sure why it could happen.

PR #267, opened 2026-04-15
Support placing code and data in CHIP memory without gcc complaining

timfel:
## Description

Allow placing both some data and some code into chip memory. Even though on the amiga (through elf2hunk) this all just is CHIP mem, gcc needs to place it in an ELF file first, and that requires separate sections for executable, read/writable, and zero-initialized things. So if one uses chip, one should be explicit, or else get linker errors when more than one kind of section is needed.

## Checklist

### DIFF
--- a/include/ace/types.h
+++ b/include/ace/types.h
@@ -48,6 +48,12 @@ typedef int32_t LONG;
 #define UNUSED_ARG __attribute__((unused))
 #define REGARG(arg, reg) arg
 #define CHIP
+// Data in chip memory. While on Amiga CHIP mem is all one region, GCC needs to tell apart executable from data, from zero-initialized data
+#define CHIP_DATA
+// Code in chip memory. While on Amiga CHIP mem is all one region, GCC needs to tell apart executable from data, from zero-initialized data
+#define CHIP_CODE
+// Zero-initialized data in chip memory. While on Amiga CHIP mem is all one region, GCC needs to tell apart executable from data, from zero-initialized data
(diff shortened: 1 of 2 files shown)


include/ace/types.h
```diff
@@ -62,6 +65,9 @@ typedef int32_t LONG;
 #define UNUSED_ARG __attribute__((unused))
 #define REGARG(arg, reg) arg
 #define CHIP __attribute__((section(".MEMF_CHIP")))
+#define CHIP_DATA __attribute__((section(".chipdata.MEMF_CHIP")))
```
timfel, line 73: These names are chosen because the bartman `ld` absorbs `.text.*` and `.data.*`, and elf2hunk looks at the string after the last `.`

tehKaiN: This might be good to have in a comment inside the code, because right now I'll remember this rationale but I'm betting I won't after a month ;)

tehKaiN (approved):
Looks good! Shouldn't it be used somewhere already, e.g. in the tilebuffer optimization code?

timfel:
> Looks good! Shouldn't it be used somewhere already, e.g. in the tilebuffer optimization code?

Done, and also added comments to each of the macros

tehKaiN (approved):

timfel:
Btw, with this and the systemDisableCaches and removing blit waits, I cut the time between tile draws in my RTS from 152 cycles to 100cycles (not counting cycles taken away by higher priority DMA) 🥳

PR #269, opened 2026-04-25
Fix font_conv docs inconsistency (-charFile param)

kwn:
This PR fixes inconsistency of `-charFile` case sensitivity in documentation.


tehKaiN (approved):
Looks good! I should probably make those switches case-insensitive anyway. Thanks for contribution!

PR #272, opened 2026-05-07
Fix BPLCON2 when enabling KILLEHB for 6 bitplanes (AGA)

kwn:
**Problem:**
For 6bpp displays, viewLoad() set bplcon2 = BV(9) (KILLEHB). That overwrote the rest of the register and cleared sprite/playfield priority bits, so hardware sprites could end up behind the playfield.

**Change:**
 Use bplcon2 = (BV(2) | BV(5) | BV(9)) so KILLEHB is set with the kickstart default values


kwn:
Let me test it properly first...

tehKaiN (approved):

PR #274, opened 2026-05-13
(partially) Fix fmode3 in scroller

kwn:
**Summary**

- Cleaned up scattered fetchmode logic by extracting it into a shared fetchmode.h. Removed duplicated code from simplebuffer.c and scrollbuffer.c
- Added (partial) FMODE 3 support to the scroller (scrollbuffer.c). This is the only approach so far that produces a working result, but it's still partially broken. The first (leftmost) tile column still scrolls incorrectly.
- Scroller artifacts were fixed in a naive way - diagonal scrolling flushes all pending margin tiles at once in tilebuffer.c, causing a noticeable slowdown

**Details**
- The fetchmode related calculations (DDFSTRT adjust, prefetch bytes, bitplane shift, BPL offset) were copy-pasted across simplebuffer.c and scrollbuffer.c with variations. This PR consolidates them into `fetchmode.h` as small static inline functions
- For the scrollbuffer specifically, FMODE 3 needed a different DDFSTRT adjustment (`fetchModeApplyScrollBufferXScrollCopper()`) to avoid a 64px ghost artifact after the vertical wrap.
- The tile buffer gained `tileBufferFlushMarginX/Y()` helpers that force-draw all remaining tiles on the opposite margin when a new margin column/row is entered. This prevents tile corruption on left/up & left/down diagonal movement but at the cost of doing all blits in one go instead of spreading them across frames

**Intent**
It's a starting point / proposal for fixing the FMODE 3 scroller, not ready to be merged yet. Ideally we should iterate and address the leftmost column glitch and find a way to avoid the full tile flush slowdown. Additionally I didn't spend much time testing it. Only run it in my game with 6 BPP lo-res, FMODES 0, 1, 3 (2 never worked).


tehKaiN (commented):
Removing code repeats and moving them to fetchmode.h looks like a good idea, perhaps it should be moved to utils - I'll think about it more.

Flushing margins is definitely a hack, so that's a no-go for merge.

Code looks good, I'll do a check for performance regressions later.

kwn:
Removed margins flush, added a proper fix.

include/ace/utils/fetchmode.h
Vairn, line 1: I like that it is header-only.

src/ace/managers/viewport/tilebuffer.c
Vairn, line 1: I'll compare this to my fix, and see if there was a different issue.

Vairn (commented):
I will test this today, and if It works well, I shall merge it.

Vairn:
Looks good to me, it might be better to have the fmode code under the utils section rather than in the viewport. 
since it is more a ultility function, then a view management function.

I have tested the code, and it is working for me. 

If you move the fetchmode.h into utils, since it is a utility function to get fetchmode settings. I shall approve it.

kwn:
Thanks for the review @tehKaiN and @Vairn 

I moved the fmode.h to the utils directory.

Vairn (approved):
Tested, and Merged. Good Stuff.

PR #275, opened 2026-05-18
Fix mini_std fwrite/fread performance issue

kwn:
`fread` and `fwrite` in `mini_std/stdio_file.c` looped `Count` times calling `Read(Size)` / `Write(Size)`. This made `disk_file.c`'s direct read path (fread(buf, ulSize, 1, f)) issue one Read(1) per byte killing load times.

Fixed both to do a single `Read(Size * Count) / Write(Size * Count)` call, matching standard C return semantics (`bytesRead / Size`).

Also swapped `disk_file.c` line 108 to `fread(pDestBytes, 1, ulSize, ...)` so the return value correctly reflects bytes read.

Tested with Bartman toolchain. Please verify with bebbo to make sure nothing breaks there.

### DIFF
--- a/src/ace/utils/disk_file.c
+++ b/src/ace/utils/disk_file.c
@@ -105,7 +105,7 @@ DISKFILE_PRIVATE ULONG diskFileRead(void *pData, void *pDest, ULONG ulSize) {
 		if(!pDiskFileData->isUninterrupted) {
 			fileAccessEnable();
 		}
-		ULONG ulReadPartSize = fread(pDestBytes, ulSize, 1, pDiskFileData->pFileHandle);
+		ULONG ulReadPartSize = fread(pDestBytes, 1, ulSize, pDiskFileData->pFileHandle);
 		pDestBytes += ulReadPartSize;
 		ulReadCount += ulReadPartSize;
 		ulSize -= ulReadPartSize;
--- a/src/mini_std/stdio_file.c
+++ b/src/mini_std/stdio_file.c
@@ -36,19 +36,20 @@ FILE *fopen(const char *restrict szFileName, const char *restrict szMode) {
 
 size_t fread(void *restrict pBuffer, size_t Size, size_t Count, FILE *restrict pStream) {
 	// http://amigadev.elowar.com/read/ADCD_2.1/Includes_and_Autodocs_3._guide/node01A0.html
-	LONG lBytesRead;
 	if(Size == 0 || Count == 0) return 0;
-	lBytesRead = Read((BPTR)pStream, pBuffer, Size * Count);
+	if(Count > SIZE_MAX / Size) return 0;
+	// Read() returns -1 on error (IoErr() has details)
+	LONG lBytesRead = Read((BPTR)pStream, pBuffer, Size * Count);
 	if(lBytesRead <= 0) return 0;
 	return (size_t)lBytesRead / Size;
 }
 
 size_t fwrite(const void *restrict pBuffer, size_t Size, size_t Count, FILE *restrict pStream) {
 	// http://amigadev.elowar.com/read/ADCD_2.1/Includes_and_Autodocs_3._guide/node01D1.html
-	LONG lBytesWritten;
 	if(Size == 0 || Count == 0) return 0;
-	lBytesWritten = Write((BPTR)pStream, pBuffer, Size * Count);
-	if(lBytesWritten <= 0) return 0;
+	if(Count > SIZE_MAX / Size) return 0;
+	LONG lBytesWritten = Write((BPTR)pStream, (void *)pBuffer, Size * Count);
+	if(lBytesWritten < 0) return 0;
 	return (size_t)lBytesWritten / Size;
 }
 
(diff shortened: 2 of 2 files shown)


src/ace/utils/disk_file.c
Vairn, line 1: yeah, the first one would return 1, while the second one should return the amount read.

Since this is a buffered read system v1 is broken, while v2 would behave as required.

But it looks like the fread/fwrite don't work the standard way. anthow, the new commit is closer to the std tho

src/mini_std/stdio_file.c
Vairn, line 1: hmm, if you want to to match the c std, I think it would be closer to this.
```c
size_t fread(void *restrict pBuffer, size_t Size, size_t Count, FILE *restrict pStream) {
    if(Size == 0 || Count == 0) return 0;

    // Check for overflow before multiplying
    if(Count > SIZE_MAX / Size) return 0;

    LONG lBytesRead = Read((BPTR)pStream, pBuffer, Size * Count);

    if(lBytesRead < 0) {
        // Read() returns -1 on error (IoErr() has details)
        // fread should set stream error indicator — skipped if no FILE struct
        return 0;
    }
    if(lBytesRead == 0) {
        // EOF — set EOF indicator if FILE struct supports it
        return 0;
    }

    // Return complete items only — tail partial item is "not read"
    // This matches C standard: partial final item is discarded from count
    return (size_t)lBytesRead / Size;
}

size_t fwrite(const void *restrict pBuffer, size_t Size, size_t Count, FILE *restrict pStream) {
    if(Size == 0 || Count == 0) return 0;

    if(Count > SIZE_MAX / Size) return 0;

    LONG lBytesWritten = Write((BPTR)pStream, (void *)pBuffer, Size * Count);

    if(lBytesWritten < 0) {
        // Write() returns -1 on error
        return 0;
    }

    return (size_t)lBytesWritten / Size;
}
```

Vairn (commented):
This is closer to a the std fread/fwrite, but not sure if we need 100% compatible.

src/ace/utils/disk_file.c
```diff
@@ -105,7 +105,7 @@ DISKFILE_PRIVATE ULONG diskFileRead(void *pData, void *pDest, ULONG ulSize) {
 		if(!pDiskFileData->isUninterrupted) {
 			fileAccessEnable();
 		}
-		ULONG ulReadPartSize = fread(pDestBytes, ulSize, 1, pDiskFileData->pFileHandle);
+		ULONG ulReadPartSize = fread(pDestBytes, 1, ulSize, pDiskFileData->pFileHandle);
```
tehKaiN, line 108: Good catch! I've changed it only in one place and skipped this one.

src/mini_std/stdio_file.c
```diff
@@ -36,25 +36,20 @@ FILE *fopen(const char *restrict szFileName, const char *restrict szMode) {
...
-		// FIXME: handle 0/negative vals
-		BytesRead += Read((BPTR)pStream, pByteBuffer, Size);
-		pByteBuffer += Size;
-	}
-	return BytesRead;
+	LONG lBytesRead;
+	if(Size == 0 || Count == 0) return 0;
+	lBytesRead = Read((BPTR)pStream, pBuffer, Size * Count);
+	if(lBytesRead <= 0) return 0;
+	return (size_t)lBytesRead / Size;
```
tehKaiN, line 55: That's definitely better! Could be improved with @Vairn suggestions though.

tehKaiN (approved):
Great job!

src/mini_std/stdio_file.c
kwn: Ok, let me give it a go

src/mini_std/stdio_file.c
kwn: Hey @Vairn . Thanks for the feedback and pointing out shortcomings. 

I updated the code. I persisted comments in my brain, but removed them from the code (except one) to make it aligned with the minimalistic style of ACE.

src/mini_std/stdio_file.c
kwn: Hey @tehKaiN. I updated the code according to @Vairn's suggestions.

src/mini_std/stdio_file.c
```diff
@@ -36,25 +37,21 @@ FILE *fopen(const char *restrict szFileName, const char *restrict szMode) {
...
 	// http://amigadev.elowar.com/read/ADCD_2.1/Includes_and_Autodocs_3._guide/node01A0.html
-	unsigned char *pByteBuffer = (unsigned char *)pBuffer;
-	size_t BytesRead = 0;
-	while(Count--) {
-		// FIXME: handle 0/negative vals
-		BytesRead += Read((BPTR)pStream, pByteBuffer, Size);
-		pByteBuffer += Size;
-	}
-	return BytesRead;
+	if(Size == 0 || Count == 0) return 0;
```
tehKaiN: please use braces for instructions inside `if` at all times, with each of them being in new line

tehKaiN (changes requested):
One thing that I didn't catch earlier. Please change that and we should be good to go.

Other than that, I've tested it on Bebbo compiler and it works nicely.

src/mini_std/stdio_file.c
kwn: Apologies, I forgot about the styling guide again. Will remember next time!

PR #276, opened 2026-05-21
You can now pass around the same buffer to different views, to reuse, or save on allocating/deallocatting ram when going from simplebuffers to tile/scrollbuffers

Vairn:
## Description

Adds optional externally supplied CHIP bitmaps for simple/scroll/tile viewport managers so buffers can be pre-allocated and reused. Includes `scrollBufferGetBitmapDimensions()` for sizing reusable memory.

Also drops legacy **v1** `.plt` loading (v2 only; v1 logs an error), fixes the unused `uwBoundHeight` warning in `scrollBufferGetBitmapDimensions`, and documents that bitmap height is driven by bound **width**, not bound height.

## Checklist


src/ace/utils/bitmap.c
```diff
@@ -41,6 +41,102 @@ static void bitmapFreeChipAligned(void *pMem, ULONG ulSize) {
...
+		pBitMap->Planes[0] = (PLANEPTR)pMem;
+		for(i = 1; i < ubDepth; ++i) {
+			pBitMap->Planes[i] = &pBitMap->Planes[i - 1][ulPlaneSize];
+		}
+	}
+	else {
+		ULONG ulPlaneSize = (ULONG)uwBytesPerRow * uwHeight;
+		for(i = 0; i != ubDepth; ++i) {
+			pBitMap->Planes[i] = (PLANEPTR)pMem + i * ulPlaneSize;
+		}
```
tehKaiN: so that's basically contiguous mode, right? So this `else {}` can be safely removed and `else if(ubFlags & BMF_CONTIGUOUS)` can be the new `else {}`.

Alternatively, we could have an error message for non-interleaved & non-contiguous scenario since this scenario isn't supported in this fn without passing an array.

Or, we could make the fn reinterpret the pMem to an array of bitplanes in this case, but that seems error-prone and we probably shouldn't do such stuff.

src/ace/utils/bitmap.c
```diff
@@ -41,6 +41,102 @@ static void bitmapFreeChipAligned(void *pMem, ULONG ulSize) {
...
+	UWORD uwBytesPerRow = uwWidth / 8;
+	ULONG ulPlaneBytes = (ULONG)uwBytesPerRow * uwHeight;
+
+	if(ubFlags & BMF_INTERLEAVED) {
+		return ulPlaneBytes * ubDepth;
+	}
+	if(ubFlags & BMF_CONTIGUOUS) {
+		return ulPlaneBytes * ubDepth;
+	}
+	return ulPlaneBytes * ubDepth;
```
tehKaiN: so uhhh just `return ulPlaneBytes * ubDepth`?

include/ace/managers/viewport/scrollbuffer.h
```diff
@@ -112,9 +115,23 @@ void scrollBufferDestroy(tScrollBufferManager *pManager);
...
+	const tVPort *pVPort, UBYTE ubMarginWidth,
+	UWORD uwBoundWidth, UWORD uwBoundHeight,
+	UWORD *puwWidth, UWORD *puwHeight
+);
+
 void scrollBufferReset(
 	tScrollBufferManager *pManager, UBYTE ubMarginWidth,
-	UWORD uwBoundWidth, UWORD uwBoundHeight, UBYTE ubBitmapFlags, UBYTE isDblBfr
+	UWORD uwBoundWidth, UWORD uwBoundHeight, UBYTE ubBitmapFlags, UBYTE isDblBfr,
+	tBitMap *pFront, tBitMap *pBack
```
tehKaiN: how about `pCustomFront`, `pCustomBack` to reflect that those things can be safely nulled? Same for other buffer managers.

include/ace/managers/viewport/scrollbuffer.h
```diff
@@ -112,9 +115,23 @@ void scrollBufferDestroy(tScrollBufferManager *pManager);
...
+ * @brief Returns pixel dimensions of scroll buffer bitmap(s) for a VPort setup.
+ * Use with bitmapGetBufferSize() to allocate reusable CHIP memory.
+ *
+ * Bitmap height uses @p uwBoundWidth (horizontal-scroll band count), not
+ * @p uwBoundHeight; the latter only limits camera Y (see TAG_SCROLLBUFFER_BOUND_HEIGHT).
+ */
+void scrollBufferGetBitmapDimensions(
+	const tVPort *pVPort, UBYTE ubMarginWidth,
+	UWORD uwBoundWidth, UWORD uwBoundHeight,
+	UWORD *puwWidth, UWORD *puwHeight
```
tehKaiN: we don't join prefixes like that, just `pWidth`, `pHeight`

include/ace/utils/bitmap.h
```diff
@@ -54,6 +54,12 @@ typedef struct _tBitMap {
  */
 #define BMF_CONTIGUOUS (1 << 6)
 
+/**
+ * @brief Plane memory is supplied externally; bitmapDestroy() frees only the
+ * tBitMap header, not bitplane pointers.
+ */
+#define BMF_EXTERNAL (1 << 7)
```
tehKaiN, line 61: I had a thought of renaming to `BMF_EXTERNAL_PLANES` but I'm not sure about it. Current one should be good enough though.

tehKaiN (changes requested):
minor stuff here and there

tehKaiN (approved):

PR #277, opened 2026-05-25
Fix hires buffers and showcase demos

Vairn:
## Description
Fix Hires screen issues, as shown in showcase.
Attempted to fix Reusable Buffer example, couldn't replicate but tightened up the demo.

## Checklist

### DIFF
--- a/src/ace/managers/viewport/scrollbuffer.c
+++ b/src/ace/managers/viewport/scrollbuffer.c
@@ -71,7 +71,7 @@ tScrollBufferManager *scrollBufferCreate(void *pTags, ...) {
 	pCopList = pVPort->pView->pCopList;
 	if(pCopList->ubMode == COPPER_MODE_BLOCK) {
 		pManager->pStartBlock = copBlockCreate(
-			pVPort->pView->pCopList, 2 * pVPort->ubBpp + 8,
+			pVPort->pView->pCopList, 2 * pVPort->ubBpp + 9,
 			// Vertically addition from DiWStrt, horizontally just so that 6bpp can be set up.
 			// First to set are ddf, modulos & shift so they are changed during fetch.
 			s_pCopperWaitXByBitplanes[pVPort->ubBpp], pVPort->uwOffsY + pVPort->pView->ubPosY -1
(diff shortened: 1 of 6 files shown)


src/ace/managers/viewport/scrollbuffer.c
```diff
@@ -153,15 +153,16 @@ tScrollBufferManager *scrollBufferCreate(void *pTags, ...) {
...
+	if(isOwnFront) {
 		if(pManager->pFront) {
 			bitmapDestroy(pManager->pFront);
 		}
 		pManager->ubFlags &= ~SCROLLBUFFER_FLAG_OWN_FRONT;
 	}
 	if(
 		(pManager->ubFlags & SCROLLBUFFER_FLAG_OWN_BACK) &&
-		pManager->pBack && pManager->pBack != pManager->pFront
+		pManager->pBack && (!isOwnFront || pManager->pBack != pManager->pFront)
```
tehKaiN, line 165: this code is a bit cryptic here - just checking for OWN_BACK and non-same buffers should be enough

src/ace/utils/extview.c
```diff
@@ -281,6 +281,9 @@ void viewLoad(tView *pView) {
 		g_pCustom->fmode = 0; // AGA fix
 		g_pCustom->bplcon3 = 0; // AGA fix
 #endif
+		if(pView->pFirstVPort->eFlags & VP_FLAG_HIRES) {
+			g_pCustom->bplcon0 |= BV(15);
```
tehKaiN, line 285: this won't work, bplcon0 is write-only

src/ace/managers/viewport/simplebuffer.c
```diff
@@ -103,6 +104,12 @@ static void simpleBufferInitializeCopperList(
 	else {
 		tCopBlock *pBlock = pManager->pCopBlock;
 		pBlock->uwCurrCount = 0; // Rewind to beginning
+		if(ubHiresCop) {
+			copMove(
```
tehKaiN, line 108: I don't like conditional length of the copperlist when the fn for getting cop length isn't conditional.

Also, why's that only for simplebuffer and only for block mode?

tehKaiN (changes requested):

kwn:
Hey @tehKaiN and @Vairn 

I fixed HiRes buffers in https://github.com/AmigaPorts/ACE/pull/280. You can consider removing the HiRes fix related code from this PR.

Vairn:
No longer needed due to being fixed in a different pull request #280.

PR #278, opened 2026-05-27
Add misc_resource and serial logging

tehKaiN:
## Description

Generalizes access to misc.resource, moves parallel-related code from joy code, adds serial-related code.

Uses this new code to add serial logging via log.h functions when ACE_DEBUG_SERIAL is enabled. Baud rate of 115200 is hardcoded.

## Motivation and Context

Allows more reasonable debugging on real hw than emitting log files. Also paves way for serial-link multiplayer games.

## ACE inclusion self-assessment

**Are there significantly different solutions to same problem? If so, is it the most common/generic one?**

It's generalization of code from joy.c, so that prevents code duplicates in case anyone needs to access serial/parallel port. Logging support is straightforward and I don't see different approaches. 

**Are other solutions better and should ACE incorporate it instead of this change in the long run?**

Not to my knowledge

**Is given functionality needed or nice to have for most of games? Is it used often in tandem with ACE?**

It's a lifesaver for real hw debugging, gameplay-wise irrelevant to most of games

**Is it small/simple enough to not require much maintenance?**

Yes

**Is the functionality related to a core Amiga concept?**

Yes

**Is it entirely, or its ACE interface, written by core ACE team or frequent contributors?**

Yes

**Is it a building block or dependency for other parts of code? How much coupling there is between this part and rest of ACE code?**

A bit of coupling between log.c and misc_resource.c, can be disabled with build switch.

**Does it allow making significantly better games at reasonable performance cost?**

Yes, multiplayer here we go!

**Is most of the code first-party?**

Yes

**Does it require additions/changes/removal of some other parts of ACE functionality/tooling for its use?**

Refactored joy.c to deduplicate code

## How Has This Been Tested?

Using two usb-uart dongles - one listened to in RealTerm, the other configured as serial output in WinUAE.

## Types of changes

## Checklist
- [ ] My change requires a change to existing documentation.
- [x] My change requires a new chapter in documentation's tutorial. - **probably**


src/ace/managers/log.c
```diff
@@ -19,20 +20,32 @@ static char s_szMsg[1024];
...
-#define uaeWrite(x)
+#define printUae(x) do {} while(0)
+#endif
+
+#if defined(ACE_DEBUG_SERIAL)
+static void printSerial(const char *szMsg) {
+	const char *pChar = szMsg;
+	while(*pChar) {
+		while((g_pCustom->serdatr & SERDATRF_TBE) == 0) continue;
+		g_pCustom->serdat = *(pChar++);
```
timfel:
```suggestion
		g_pCustom->serdat = ((UWORD)1 << 8)|*(pChar++);
```

timfel (approved):

PR #280, opened 2026-06-05
Fix scroller - 6BPP & 7BPP, fix HiRes

kwn:
## Description

Fix scroll buffer display setup for higher bit depths and hires modes.

This PR improves scroll buffer behavior for 6bpp and 7bpp in FMODE 0, 1, and 3. FMODE 3 is usable but still shows minor horizontal black-line glitches.

It also fixes hires mode setup by correctly composing and writing `BPLCON0`, and simplifies the related `BPLCON0` and `BPLCON2` logic in `src/ace/utils/extview.c`.

Known limitation: 8bpp scroll buffer modes are still not fixed. They likely need a deeper change to the scrolling/wrapping mechanism because the current Copper update approach appears to exceed available timing.

## Checklist

### DIFF
--- a/src/ace/managers/viewport/simplebuffer.c
+++ b/src/ace/managers/viewport/simplebuffer.c
@@ -7,7 +7,6 @@
 #include <ace/utils/tag.h>
 #include <ace/utils/extview.h>
 #include <ace/utils/fetchmode.h>
-#include <ace/generic/screen.h> // Has the look up table for the COPPER_X_WAIT values.
 #ifdef AMIGA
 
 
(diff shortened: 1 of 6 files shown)


src/ace/managers/viewport/simplebuffer.c
```diff
@@ -42,13 +41,8 @@ static void simpleBufferInitializeCopperList(
 	UWORD uwModulo = pManager->pFront->BytesPerRow - (pVPort->uwWidth >> 3);
 
 	// http://amigadev.elowar.com/read/ADCD_2.1/Hardware_Manual_guide/node0085.html
```
tehKaiN: this comment should go `fetchModeGetDDfStrt()` and `fetchModeGetDDfStop()` since it describes the calcs

tehKaiN (commented):
Generally looks good.

I'm not sure whether those extended aga-related ifs/switches in fetchmode.h won't introduce extra overhead for ocs since there's no e.g. ifdef aga in calculating bitplane shift, but if that happens we'll add hot paths.

PR #281, opened 2026-06-05
Fix double-buffer init after reusable bitmap support

kwn:
## Description
Fix double-buffer initialization in `simplebuffer` after reusable bitmap support.

When `TAG_SIMPLEBUFFER_IS_DBLBUF` was enabled without custom front/back bitmaps, the back buffer could remain `NULL`, so a real second buffer was not created. This caused flickering during redraws. The fix ensures a back buffer is allocated when double buffering is requested and no custom back buffer is provided.

## Checklist


NZjeux26:
Adding here since it was discussed on the AmigaGameDev discord. I have had a double buffer flicker appear after the most recent ace update and checking out this pull request and testing on macACE on my machine seems to have resolved the problem.

tehKaiN (approved):
Looks good, thanks! And thanks @NZjeux26 for tests!

I'm not completely happy on having two different `pBack` states to check here for, but that would need a further refactor of code and is not a subject of this fix.

PR #282, opened 2026-07-21
Add the ability to get pak file by their hash

steamknight:
## Description
Pak files are awesome, but recompute the path hash every time you open a file.

Added a way to skip the adler32 hashing on every file read if you already know the file hash. Also added a way to get that hash if you don't know it.

Lastly, refactored the original function to avoid code duplication.

Technically, it's possible to go further and have a `pakFileGetFileByIndex` with a public `pakFileGetFileIndex`, but I didn't want to suggest too many changes in this PR.

I wasn't sure what the TODO comment next to the adler32buffer call was about, so I kept it.

## Checklist

### DIFF
--- a/include/ace/utils/pak_file.h
+++ b/include/ace/utils/pak_file.h
@@ -36,6 +36,8 @@ tFile *pakFileGetFileByPath(tPakFile *pPakFile, const char *szInternalPath);
 
 tFile *pakFileGetFileByHash(tPakFile *pPakFile, ULONG ulPathHash);
 
+tFile *pakFileGetFileByIndex(tPakFile *pPakFile, UWORD uwIndex);
+
 ULONG pakFileGetPathHash(const char *szPath);
 
 #endif
--- a/include/ace/utils/pak_file.h
+++ b/include/ace/utils/pak_file.h
@@ -32,7 +32,7 @@ tPakFile *pakFileOpen(const char *szPath, UBYTE isUninterrupted);
 
 void pakFileClose(tPakFile *pPakFile);
 
-tFile *pakFileGetFile(tPakFile *pPakFile, const char *szInternalPath);
+tFile *pakFileGetFileByPath(tPakFile *pPakFile, const char *szInternalPath);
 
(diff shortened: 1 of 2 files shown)


steamknight:
Mmm... Not sure why the CI failed in zero seconds. Unfortunately, it doesn't say why it failed.

Vairn:
> Mmm... Not sure why the CI failed in zero seconds. Unfortunately, it doesn't say why it failed.

Something weird happens with Jenkins, I get it too. So just don't worry for now.

Vairn (approved):
Looks good to me.

tehKaiN (approved):
Looks good! I'd probably change `pakFileGetFile()` to `pakFileGetFileByPath()` for parity with `pakFileGetFileByHash()`... or maybe not since that name is ugly as hell. ;)

Also, I do wonder if accessing by hash is better - it theoretically allows you to preserve them across builds, but is that a valid scenario where e.g. asset path might change? If it's only for runtime and reusing during the same game run, then perhaps it would be better/faster to calculate and reuse file index instead?

On the other hand, using hashes allows doing fancy stuff like overlay paks and loading 3rd party mods, so perhaps both things are worth considering.

I generally approve the changes, but before we merge this thing, lemme know if your use case wouldn't benefit more from reusing the index.

Re CI - it acts a bit weird when PRs of ppl outside of ACE team trigger it.

Vairn:
I have renamed the two functions for you.
Parity is good, it lets you know that you are looking by Path or by Hash.

steamknight:
For my use case, I have a custom editor I use to author a game file that gets packaged along with the other assets. The game file holds the hashes (computed by my editor), so when loading, it can reference the files in the pak.

However, since you gave us the ability to use a manifest to set the order of the files in the pak, I could just store the index instead if we had a `pakFileGetFileByIndex` or something similar, thus avoiding the lookup altogether. I'd forgotten that was an option.

tehKaiN:
Will you be willing to add that with this PR next to byHash() fn, or should I add that for you at a bit later time?

steamknight:
I can do it!

tehKaiN (approved):

tehKaiN:
Thanks!

